PR #4, opened 2019-02-24
Add helper scripts

PeterJCLaw:
Some helper scripts to simplify development. Also for tab completion.

### DIFF
--- a/scripts/install-for-development
+++ b/scripts/install-for-development
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+if [ -z "$VIRTUAL_ENV" ]
+then
+    echo "Refusing to run outside a virtual environment. If you already have"
+    echo "pipenv configured globally you should use it directly."
+    exit 1
+fi
+
+pip install pipenv
+PIPENV_VERBOSITY=-1 pipenv install --dev
--- a/scripts/lint
+++ b/scripts/lint
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+PIPENV_VERBOSITY=-1 exec pipenv run lint-yaml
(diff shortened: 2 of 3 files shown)


RealOrangeOne:
I'm not sure I see the merit of these scripts, especially the serve one. Forcing the commands to be run through `pipenv` ensures there's just 1 right way to do something.

The development install one should be covered by the readme?

PeterJCLaw:
The purpose of these scripts is to roughly follow https://github.com/github/scripts-to-rule-them-all, meaning that the user doesn't need to worry too much about the underlying technologies in order to get the docs up & running.

I agree the install one still needs some bootstrapping that regard, but it feels more useful to me than otherwise.

Importantly if we all actually use these scripts then their regular use ensures that they can't get out of date, which is more than one could ever say for a README.

They also have the added benefit of aligning with the muscle memory which people will have elsewhere and automatically tab-completing in pretty much any environment (AFAIK `pipenv` has no tab completion anywhere).

Ideally I'd also like to see any linting we set up also be able to integrate into editors.

I managed to completely overlook the yaml linting while preparing #5, I think partly as it's in a README section called "tests" which I would have assumed meant tests for any software, rather than checks against the spelling of the content. (I think of the latter as "linting"). Maybe the solution to this specific case is to tweak the README?

trickeydan:
Given that this repository uses `pipenv`. I would strongly suggest either moving these scripts into the `[scripts]` section of the `Pipfile` OR at the very least adding them such that execution is easier for those using `pipenv`.

e.g
```
[scripts]

"lint" = "./scripts/lint"
```

PR #16, opened 2019-03-17
Update kit handback plan

RealOrangeOne:
Fixes https://github.com/srobo/tasks/issues/115

I've intentionally _not_ encoded any location information for this years competition, so the documentation can remain year-agnostic. 

Primary changes:
- Reflow content. Headers are slightly nicer to visually grep than a single list
- Add special case for those leaving before end of competition.

## TODO
- [x] Find out where the related documents are
- [x] Find out where teams should ship their kits to after the competition (https://github.com/srobo/tasks/issues/186)

### DIFF
--- a/docs/competition/teams/role-descriptions/kit-return.md
+++ b/docs/competition/teams/role-descriptions/kit-return.md
@@ -49,7 +49,7 @@ This workflow is shown in the diagram below.
 
 ![Kit return flow](../diagrams/kit-return-desk-flow.svg)
 
-For each returned kit:
+## For each returned kit
 
 * Remove any batteries from the kit and store them separately
 * Seal the handles of the kit with tape
--- a/docs/competition/teams/role-descriptions/kit-return.md
+++ b/docs/competition/teams/role-descriptions/kit-return.md
@@ -32,7 +32,7 @@ For each returned kit:
   Make all reasonable attempts to retrieve kit there and then i.e. if they've left a bit upstairs they go and get it.
 * If they do not have something, record it on a missing kit form.
 * If the kit return form indicates that they have more than a normal kit to return (i.e. they had broken kit and did not return it upon arrival), see if they can return it now. If not, record it on a missing kit form.
-* Indicate if all parts are present or if a missing kit form has been filled in on the kit return form (provided by me).
+* Indicate if all parts are present or if a missing kit form has been filled in on the kit return form.
 * Write down the part code of the Really Useful Box on the kit return form.
 * Get their team leader to sign and write their name on the kit return form.
 
--- a/docs/competition/teams/role-descriptions/kit-return.md
+++ b/docs/competition/teams/role-descriptions/kit-return.md
@@ -19,12 +19,6 @@ rejected. When presented with a RUB full of kit, kit return staff should:
 
 **Don't let people leave the competition for good with any kit if they aren't on the list of teams authorised to keep their kit**
 
-For each returned kit:
-
-* Remove any batteries from the kit and store them separately
-* Seal the handles of the kit with tape
-* Stick a post-it on the end of the box with the team's TLA on it (to allow for double checking later on)
-* Stack the box with the post-it visible
 
 ## Teams returning kit
 
@@ -54,6 +48,14 @@ Teams leaving early should follow the same process as above, though we should pu
 This workflow is shown in the diagram below.
 
 ![Kit return flow](../diagrams/kit-return-desk-flow.svg)
+
+For each returned kit:
+
+* Remove any batteries from the kit and store them separately
+* Seal the handles of the kit with tape
+* Stick a post-it on the end of the box with the team's TLA on it (to allow for double checking later on)
+* Stack the box with the post-it visible
+
 
 ## Required Items
 You should have the following items present for your duties. If you do not, please let the competition team coordinator know as soon as possible. The documents can be found in the [`kit-coordination-documents`](https://github.com/srobo/kit-coordination-documents) repo.
--- a/docs/competition/teams/role-descriptions/kit-return.md
+++ b/docs/competition/teams/role-descriptions/kit-return.md
@@ -36,6 +36,10 @@ For each returned kit:
 * Write down the part code of the Really Useful Box on the kit return form.
 * Get their team leader to sign and write their name on the kit return form.
 
+### Teams leaving before competition-end
+
+If a team wishes to leave before the end of the competition, we should attempt to find any missing kit, rather than recording it on a missing kit form.
+
 ## Teams keeping kit
 
 * Write down the part code of the Really Useful Box on the kit return form.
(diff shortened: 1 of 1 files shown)


docs/competition/teams/role-descriptions/kit-return.md
```diff
@@ -19,35 +19,42 @@ rejected. When presented with a RUB full of kit, kit return staff should:
...
+* Stick a post-it on the end of the box with the team's TLA on it (to allow for double checking later on)
+* Stack the box with the post-it visible
+
+## Teams returning kit
+
+* Check the contents against a list of parts expected in the kit.
+  Make all reasonable attempts to retrieve kit there and then i.e. if they've left a bit upstairs they go and get it.
+* If they do not have something, record it on a missing kit form.
+* If the kit return form indicates that they have more than a normal kit to return (i.e. they had broken kit and did not return it upon arrival), see if they can return it now. If not, record it on a missing kit form.
+* Indicate if all parts are present or if a missing kit form has been filled in on the kit return form (provided by me).
```
PeterJCLaw: Who is the "me" here? (I suspect it was originally Rich, but we should update this now)

docs/competition/teams/role-descriptions/kit-return.md
```diff
@@ -19,35 +19,42 @@ rejected. When presented with a RUB full of kit, kit return staff should:
...
+* Write down the part code of the Really Useful Box on the kit return form.
+* Get their team leader to sign and write their name on the kit return form.
+
+### Teams leaving before competition-end
+
+If a team wishes to leave before the end of the competition, we should attempt to find any missing kit, rather than recording it on a missing kit form.
```
PeterJCLaw: Is there a way we can frame this which is clearer about the intent here? It's not that we'll prevent them leaving at all unless they return the kit (they might not have brought it with them), but rather that we'll spend longer getting them to search for things they might have left around the venue.

Perhaps something like:
```suggestion
Teams leaving early should follow the same process as above, though we should push them to spend more time searching for any missing items they might have lost at the venue. If they are unable to find the items or are sure that they didn't bring them, then we should record the item on a missing kit form (as above).
```

docs/competition/teams/role-descriptions/kit-return.md
```diff
@@ -19,35 +19,42 @@ rejected. When presented with a RUB full of kit, kit return staff should:
...
+* Stick the appropriate 'authorised to keep kit' sheet onto the kit of the Really Useful Box with tape.
+* Kit should be shipped to the postal address listed on [the website](https://studentrobotics.org/contact/).
 
 This workflow is shown in the diagram below.
 
 ![Kit return flow](../diagrams/kit-return-desk-flow.svg)
 
 ## Required Items
-You should have the following items present for your duties. If you do not, please let the competition team coordinator know as soon as possible.
+You should have the following items present for your duties. If you do not, please let the competition team coordinator know as soon as possible. The documents can be found in the [`kit-coordination-docs`](https://github.com/srobo/kit-coordination-docs) repo.
```
PeterJCLaw: We should assume that volunteers at the competition have no access to other resources. I'm not sure what the best way to frame that here is, however we should be clear that at the competition any needed paperwork will already be provided and that the link is a reference for the online version of these docs.

docs/competition/teams/role-descriptions/kit-return.md
```diff
@@ -19,35 +19,42 @@ rejected. When presented with a RUB full of kit, kit return staff should:
...
+If a team wishes to leave before the end of the competition, we should attempt to find any missing kit, rather than recording it on a missing kit form.
+
+## Teams keeping kit
+
+* Write down the part code of the Really Useful Box on the kit return form.
+* Give them two batteries, a battery charging bag and a charger+PSU. Record the part numbers of these five things on the kit return form.
+* Get their team leader to sign and write their name on the kit return form.
+* Give the team a return kit and ensure you inform them that they will be responsible for return shipping charges.
+* Stick the appropriate 'authorised to keep kit' sheet onto the kit of the Really Useful Box with tape.
+* Kit should be shipped to the postal address listed on [the website](https://studentrobotics.org/contact/).
```
PeterJCLaw: Perhaps we could include this information as part of the sheet itself? (see also https://github.com/srobo/kit-coordination-docs/commit/2454c993f028301012fe1b2edf5de465a44e0f45#r32890277)

docs/competition/teams/role-descriptions/kit-return.md
```diff
@@ -19,35 +19,42 @@ rejected. When presented with a RUB full of kit, kit return staff should:
 
 **Don't let people leave the competition for good with any kit if they aren't on the list of teams authorised to keep their kit**
```
PeterJCLaw: Could we add a list of required equipment and paperwork to this document?

RealOrangeOne:
We should document where the relevant forms will end up.

https://github.com/srobo/kit-coordination-docs/pull/4#discussion_r268449775

docs/competition/teams/role-descriptions/kit-return.md
RealOrangeOne: This already exists at the bottom of the page.

docs/competition/teams/role-descriptions/kit-return.md
RealOrangeOne: This has been done.

docs/competition/teams/role-descriptions/kit-return.md
```diff
@@ -19,35 +19,46 @@ rejected. When presented with a RUB full of kit, kit return staff should:
...
+* Get their team leader to sign and write their name on the kit return form.
+* Give the team a return kit and ensure you inform them that they will be responsible for return shipping charges.
+* Stick the appropriate 'authorised to keep kit' sheet onto the kit of the Really Useful Box with tape.
+* Kit should be shipped to the postal address listed on [the website](https://studentrobotics.org/contact/).
 
 This workflow is shown in the diagram below.
 
 ![Kit return flow](../diagrams/kit-return-desk-flow.svg)
 
+For each returned kit:
```
PeterJCLaw: Would this make more sense in the "Teams returning kit" section (rather than the "Teams keeping kit" section)?

PeterJCLaw: I'm thinking before the "Teams leaving before competition-end" heading.

PeterJCLaw: Alternatively, make this a heading?

docs/competition/teams/role-descriptions/kit-return.md
```diff
@@ -19,35 +19,46 @@ rejected. When presented with a RUB full of kit, kit return staff should:
...
+Teams leaving early should follow the same process as above, though we should push them to spend more time searching for any missing items they might have lost at the venue. If they are unable to find the items or are sure that they didn't bring them, then we should record the item on a missing kit form (as above).
+
+## Teams keeping kit
+
+* Write down the part code of the Really Useful Box on the kit return form.
+* Give them two batteries, a battery charging bag and a charger+PSU. Record the part numbers of these five things on the kit return form.
+* Get their team leader to sign and write their name on the kit return form.
+* Give the team a return kit and ensure you inform them that they will be responsible for return shipping charges.
+* Stick the appropriate 'authorised to keep kit' sheet onto the kit of the Really Useful Box with tape.
+* Kit should be shipped to the postal address listed on [the website](https://studentrobotics.org/contact/).
```
PeterJCLaw: Maybe clarify the website link since this will be printed?
```suggestion
* Kit should be shipped to the postal address listed on the website at https://studentrobotics.org/contact/, which is also included on the kit-authorisation form.
```

PeterJCLaw: (also above)

PR #19, opened 2019-03-23
Add information about Tinker Time

sedders123:
This is an attempt at documenting how Tinker Time works at a competition. This is based on how we have run things in the past, with some small tweaks

### DIFF
--- a/docs/competition/team-support/tinker-time.md
+++ b/docs/competition/team-support/tinker-time.md
@@ -1,0 +1,13 @@
+# Tinker Time
+
+Tinker Time is time allocated in the schedule to allow teams to test their robot in the arena outside of a match.
+
+During Tinker Time up to 3 members of a team may enter the arena.
+
+To book Tinker Time teams must find the Tinker Time Coordinator, stationed by a whiteboard near the staging area, who will allocate teams slots based on what's available.
+
+Each 'slot' will last 10 minutes, 4 teams can be in the arena at once (one in each corner), robots must enter the arena through staging. However during Tinker Time the second door will be opened in scafolding and non robot carrying comeptitors may enter and exit through here.
+
+Tinker Time starts directly after the welcome talk, however this means some teams will not be saftey tested yet. In lieu of a high vis, the Badgers in staging can make a call on saftey looking for obvious issues and using common sense. The Badgers should err on the side of caution and if a team is rejected by them they must get approval from the Health and Safety coordinator before entering the arena.
+
+On Saturday teams can book a limited number of slots directly after the welcome talk. On the Sunday teams can find the Tinker Time Coordinator and book a slot if one is available.
--- a/docs/competition/team-support/tinker-time.md
+++ b/docs/competition/team-support/tinker-time.md
@@ -8,6 +8,6 @@ To book Tinker Time teams must find the Tinker Time Coordinator, stationed by a 
 
 Each 'slot' will last 10 minutes, 4 teams can be in the arena at once (one in each corner), robots must enter the arena through staging. For SR2019, where we have a scaffold-based arena, during Tinker Time the second door will be opened and non-robot-carrying competitors may enter and exit through here.
 
-Tinker Time starts directly after the welcome talk, however this means some teams will not be saftey tested yet. In lieu of a high vis, the Badgers in staging can make a call on saftey looking for obvious issues and using common sense. The Badgers should err on the side of caution and if a team is rejected by them they must get approval from the Health and Safety coordinator before entering the arena.
+Tinker Time starts directly after the welcome talk, however this means some teams will not be safety tested yet. In lieu of a high vis, the Badgers in staging can make a call on safety looking for obvious issues and using common sense. The Badgers should err on the side of caution and if a team is rejected by them they must get approval from the Health and Safety coordinator before entering the arena.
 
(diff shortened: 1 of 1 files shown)


docs/competition/team-support/tinker-time.md
```diff
@@ -0,0 +1,13 @@
...
+
+Tinker Time is time allocated in the schedule to allow teams to test their robot in the arena outside of a match.
+
+During Tinker Time up to 3 members of a team may enter the arena.
+
+To book Tinker Time teams must find the Tinker Time Coordinator, stationed by a whiteboard near the staging area, who will allocate teams slots based on what's available.
+
+Each 'slot' will last 10 minutes, 4 teams can be in the arena at once (one in each corner), robots must enter the arena through staging. However during Tinker Time the second door will be opened in scafolding and non robot carrying comeptitors may enter and exit through here.
+
+Tinker Time starts directly after the welcome talk, however this means some teams will not be saftey tested yet. In lieu of a high vis, the Badgers in staging can make a call on saftey looking for obvious issues and using common sense. The Badgers should err on the side of caution and if a team is rejected by them they must get approval from the Health and Safety coordinator before entering the arena.
```
trickeydan:
```suggestion
Tinker Time starts directly after the welcome talk, however this means some teams will not be safety tested yet. In lieu of a high vis, the Badgers in staging can make a call on safety looking for obvious issues and using common sense. The Badgers should err on the side of caution and if a team is rejected by them they must get approval from the Health and Safety coordinator before entering the arena.
```

docs/competition/team-support/tinker-time.md
```diff
@@ -0,0 +1,13 @@
+# Tinker Time
+
+Tinker Time is time allocated in the schedule to allow teams to test their robot in the arena outside of a match.
+
+During Tinker Time up to 3 members of a team may enter the arena.
+
+To book Tinker Time teams must find the Tinker Time Coordinator, stationed by a whiteboard near the staging area, who will allocate teams slots based on what's available.
+
+Each 'slot' will last 10 minutes, 4 teams can be in the arena at once (one in each corner), robots must enter the arena through staging. However during Tinker Time the second door will be opened in scafolding and non robot carrying comeptitors may enter and exit through here.
```
PeterJCLaw: Please could we note explicitly that the comment about the second door is specific to 2019?

Perhaps something like:
```suggestion
Each 'slot' will last 10 minutes, 4 teams can be in the arena at once (one in each corner), robots must enter the arena through staging. For SR2019, where we have a scaffold-based arena, during Tinker Time the second door will be opened and non-robot-carrying competitors may enter and exit through here.
```

docs/competition/team-support/tinker-time.md
PeterJCLaw: I'm not sure we've had in-arena tinker times before the first leagues in the past, not least because the arena is often not ready.

Either way, I'm not sure I like the idea of non-safety-checked robots being allowed in the arena. At best it means a bad precedent ("but we were allowed earlier") and at worst it means we're putting other competitors at risk.
Was this aspect discussed with @Scarzy at all?

RealOrangeOne: > Was this aspect discussed with @Scarzy at all?

It was not. We did discuss it as a group on Saturday, and this was the best idea we could come up with, outside of pushing Tinker time back a while to allow testing to happen of the first slot, and forcing people to go to the arena early to get their robots tested. It also means if for whatever reason a team fails a test, they don't miss a tinker time slot, which gets wasted.

sedders123: If a team fails the health and safety test they should not be allowed in the arena. While tinker time is incredibly useful to teams we should certainly not put it ahead of safety. 

If we haven't done pre-league tinker time at other competitions is there a reason we're doing it at this one? The problem of robots not having time to be tested goes away if we push tinker time back so it doesn't start immediately after the welcome talk. Is it possible for us to do this?

Scarzy: I'm in 2 minds over this.

I'm not convinced it puts the competitors at any more of a risk than if the robot in the pits next to them isn't safe. I actually think that the arena during Tinker Time is probably safer than the pits, as it's constantly under observation and there will be fewer people moving about in the immediate vicinity. I also think it will be rather restrictive on doing the safety checks if we have to not only do them with priority for who's got an upcoming match, but also for who's got an upcoming tinker time. 

> It also means if for whatever reason a team fails a test, they don't miss a tinker time slot, which gets wasted.
This I do consider to be an issue, if they have _failed_ then we shouldn't be letting them in until the issue has been addressed. I cannot stress how much I _don't_ want to be having the conversation of "Yes, we did identify that this robot could burst into flames, however we didn't want to waste a tinker time slot so let them go ahead anyway" with the police or insurance companies...

On balance, I currently think the best approach is to follow what's described, however the Badgers should contact me and check that there isn't any reason why robots without hi-vis should be kept out of the arena.

PeterJCLaw: Sounds good. Should we update the text generally to have something about the Badgers lasing with the Health & Safety lead about which robots can enter the arena? (This might fit better into their briefings than them checking in for every robot?)

I think this is the remaining thing on this PR, so it would be good to get this in.

sedders123: Have clarified this. Please check that I've understood the desired procedure correctly

PR #20, opened 2019-03-23
Add overnight battery loan docs

RealOrangeOne:
Fixes https://github.com/srobo/tasks/issues/213

## ToDo

- [x] Write the overnight loan form
- [ ] Populate link to overnight loan form (https://github.com/srobo/kit-coordination-docs/pull/6)

### DIFF
--- a/docs/competition/teams/process-docs/overnight-battery-loan.md
+++ b/docs/competition/teams/process-docs/overnight-battery-loan.md
@@ -1,0 +1,13 @@
+# Overnight Battery Loan
+
+Teams often request a battery and charger for use overnight to test their robots.
+
+## Obtaining batteries
+
+Batteries and chargers should be collected from helpdesk by teams. Teams should be given a charger, and no more than 2 batteries.
+
+The teams TLA, and the part codes of the batteries and changer should be recorded on the [Overnight battery loan]() form.
+
+## Returning batteries
+
+Batteries should be returned to reception on Sunday when teams enter the venue. The list of teams who took batteries overnight should be used to confirm which teams bring them back.
--- a/docs/competition/teams/process-docs/overnight-battery-loan.md
+++ b/docs/competition/teams/process-docs/overnight-battery-loan.md
@@ -4,7 +4,7 @@ Teams often request a battery and charger for use overnight to test their robots
 
 ## Obtaining batteries
 
-Batteries and chargers should be collected from helpdesk by teams. Teams should be given a charger, and no more than 2 batteries.
+Batteries and chargers should be collected from helpdesk by teams. Teams should be given a charger, a battery bag, and 2 batteries.
 
 The teams TLA, and the part codes of the batteries and changer should be recorded on the [Overnight battery loan]() form.
 
--- a/docs/competition/teams/process-docs/overnight-battery-loan.md
+++ b/docs/competition/teams/process-docs/overnight-battery-loan.md
@@ -6,7 +6,7 @@ Teams often request a battery and charger for use overnight to test their robots
 
 Batteries and chargers should be collected from helpdesk by teams. Teams should be given a charger, a battery bag, and 2 batteries.
 
-The teams TLA, and the part codes of the batteries and changer should be recorded on the [Overnight battery loan]() form.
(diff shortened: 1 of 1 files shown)


docs/competition/teams/process-docs/overnight-battery-loan.md
```diff
@@ -0,0 +1,13 @@
+# Overnight Battery Loan
+
+Teams often request a battery and charger for use overnight to test their robots.
+
+## Obtaining batteries
+
+Batteries and chargers should be collected from helpdesk by teams. Teams should be given a charger, and no more than 2 batteries.
```
PeterJCLaw: Also a battery bag.

I would encourage that we give teams a _fixed_ quantity of things, which will make recovering the items from them in the morning easier. Specifically, if we give everyone two batteries then we don't need to worry about whether they only had one battery or two when we collect them back - we just demand two batteries from them.

docs/competition/teams/process-docs/overnight-battery-loan.md
```diff
@@ -0,0 +1,13 @@
...
+
+The teams TLA, and the part codes of the batteries and changer should be recorded on the [Overnight battery loan]() form.
```
PeterJCLaw: missing link here?

docs/competition/teams/process-docs/overnight-battery-loan.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Returning batteries
+
+Batteries should be returned to reception on Sunday when teams enter the venue. The list of teams who took batteries overnight should be used to confirm which teams bring them back.
```
PeterJCLaw: What enforces this? Are we relying on their honesty, or is this something for reception to do? If the latter
* how does reception know which teams to collect batteries from?
* we should modify the reception docs to account for this and to ensure they have boxes for collecting the kit into on the Sunday morning

docs/competition/teams/process-docs/overnight-battery-loan.md
RealOrangeOne: Ah, yes. I'll populate it once https://github.com/srobo/kit-coordination-docs/pull/6 merges.

docs/competition/teams/process-docs/overnight-battery-loan.md
RealOrangeOne: > how does reception know which teams to collect batteries from

The list of teams who took batteries overnight should be used to confirm which teams bring them back.

> we should modify the reception docs

Agreed.

docs/competition/teams/process-docs/overnight-battery-loan.md
Adimote: @RealOrangeOne can you populate?

docs/competition/teams/process-docs/overnight-battery-loan.md
PeterJCLaw: Ah, ok. Could we be explicit about that?
``` suggestion
Batteries should be returned to reception on Sunday when teams enter the venue.
Once all the teams have left on Saturday, the list of teams who took batteries overnight should be passed to reception who will ensure teams return their batteries on the Sunday morning.
```

Hrm. If we do pass the list over to reception, how does the helpdesk then know which items the team had overnight?
Would it make sense to instead provide reception with a _separate_ list of TLAs, perhaps _copied_ from the list of loans?

PeterJCLaw: I see what you mean about the quoted sentence, however I find that very easy to read as being about the Saturday and it seems easy to overlook that it applies to the Sunday morning too.
There's also no mention in the reception docs about checking teams who've borrowed batteries overnight and directing them to the arrivals desk to return those items.

PeterJCLaw:
@RealOrangeOne what's the status of this? I suggest that if this is close to what actually happened at SR2019 we aim to merge _that_ and then worry about improving things separately.

RealOrangeOne:
Agreed. What's documented here is close enough to what happened at SR2019 to be useful for future.

I suggest we merge this, and i'll improve the content on this at a later date, well before the start of competition planning for SR2020

PR #21, opened 2019-03-23
Updates and Modifications to Charging Procedure

jcwilliamson:
Changes to charging a discharged battery to accommodate new chargers.
Updated equipment list.
Allocation of tasks to roles.
Link to battery docs added.

Original rework by Dan Trickey

### DIFF
--- a/docs/competition/team-support/battery-charging.md
+++ b/docs/competition/team-support/battery-charging.md
@@ -33,7 +33,7 @@ There are two locations involved in battery charging. The equipment required at 
 
 ## Volunteer Requirements
 
-Every volunteer performing the Battery Charging role must have read this document fully. Any doubts must be raisedwith the Health and Safety officer ASAP.
+Every volunteer performing the Battery Charging role must have read this document fully. Any doubts must be raised with the Health and Safety officer ASAP.
 
 ## Safety
 
@@ -73,7 +73,7 @@ Batteries are charged as follows:
     * if the battery is damaged, tape over the yellow XT60 connector with red tape and place it in the damaged battery box.
 1. If the battery is not damaged, plug the yellow XT60 connector into the charger cable and the white balance connector into the 4-pin socket on the side of the charger.
 1. Place the battery in the charging bag next to the charger.
-1. Hold down the Start/Enter button for about 3 seconds until the screen displays `R:3SER S:3SER CONFIRM(ENTER)`. If both numbers are not 3 then disconnect the battery and summon Rich Barlow via Helpdesk.
+1. Hold down the Start/Enter button for about 3 seconds until the screen displays `R:3SER S:3SER CONFIRM(ENTER)`. If both numbers are not 3 then disconnect the battery and summon the Health and Safety officer.
 1. If the display is as expected, press Start/Enter to commence charging.
 After about 1 hour the battery should complete charging - the screen will display a charge complete message. Simply disconnect the battery from the charger and place it in the charged battery box.
 
--- a/docs/competition/team-support/battery-charging.md
+++ b/docs/competition/team-support/battery-charging.md
@@ -19,13 +19,11 @@ There are two locations involved in battery charging:
     - 1 x LiPo Bag for damaged batteries, clearly labelled, red taped
     - At least one roll of red insulating tape
 - Battery Charging Station
-    - At least `n` battery chargers
-    - Enough LiPo charging bags for every charger.
     - 1 x Discharged Battery Box (18L RUB), clearly labelled
     - 1 x Charged Battery Box (18L RUB), clearly labelled
     - 1 x Damaged battery box (18L RUB), clearly labelled, red taped
     - At least one roll of red insulating tape
-
+    - Example good batteries (one brand new, one squidgy)
 
 ### Safe Use of Battery Charging Equipment - Battery Charging Station 
 
@@ -35,7 +33,7 @@ There are two locations involved in battery charging:
 
 ## Volunteer Requirements
 
-Every volunteer performing the Battery Charging role must have read this document fully. Any doubts must be raised before beginning the role.
+Every volunteer performing the Battery Charging role must have read this document fully. Any doubts must be raisedwith the Health and Safety officer ASAP.
 
 ## Safety
 
@@ -43,7 +41,7 @@ The biggest risk of damage or injury is from the banana plugs coming unplugged f
 
 The batteries themselves have many built in safety features and, unless severely abused, are completely safe. All batteries undergoing charging must be placed inside a fireproof charging bag. In the extremely unlikely event of a fire due to a battery DO NOT attempt to tackle the fire - the batteries contain a flammable metal (Lithium) which cannot be safely tackled with most fire extinguishers.
 
-If at any point you have any doubts about safety, immediately ask the helpdesk to summon the Health & Safety officer.
+If at any point you have any doubts about safety, immediately summon the Health & Safety officer.
 
 ## Procedures
 
@@ -51,7 +49,7 @@ If at any point you have any doubts about safety, immediately ask the helpdesk t
 
 - Teams will bring their discharged battery to the Helpdesk
 - Discharged batteries should be placed in the discharged battery box
-    - If the battery is damaged, red tape should be placed on it and it should be placed in the damaged LiPo bag instead of the discharged battery box. 
+- If the battery is damaged, red tape should be placed on it and it should be placed in the damaged LiPo bag instead of the discharged battery box. 
 - A charged battery should be retrieved from the charged battery box and handed to the team.
     - Do not allow teams to take batteries from the box themselves.
     - If multiple batteries are returned, only give out one charged battery to each team.
@@ -71,12 +69,12 @@ All of the battery chargers will be set-up and ready to use. Before a battery is
 
 Batteries are charged as follows:
 
-- Check that the battery is not physically damaged. There should be no splits or tears in the body of the battery, no nicks in the wire insulation, no exposed wire and the battery should not be so swollen that it is no longer squidgy. There will be a couple of example batteries at the Battery Charging station to compare to.
+- 1. Check that the battery is not physically damaged. There should be no splits or tears in the body of the battery, no nicks in the wire insulation, no exposed wire and the battery should not be so swollen that it is no longer squidgy. There will be a couple of example batteries at the Battery Charging station to compare to.
 - If the battery is damaged, tape over the yellow XT60 connector with red tape and place it in the damaged battery box.
 - If the battery is not damaged, plug the yellow XT60 connector into the charger cable and the white balance connector into the 4-pin socket on the side of the charger.
-- Place the battery in the charging bag next to the charger.
-- Follow the instructions as found in the [Student Robotics kit docs](https://www.studentrobotics.org/docs/kit/batteries/).
-
+- 1. Place the battery in the charging bag next to the charger.
+- 1. Hold down the Start/Enter button for about 3 seconds until the screen displays `R:3SER S:3SER CONFIRM(ENTER)`. If both numbers are not 3 then disconnect the battery and summon Rich Barlow via Helpdesk.
+- 1. If the display is as expected, press Start/Enter to commence charging.
 After about 1 hour the battery should complete charging - the screen will display a charge complete message. Simply disconnect the battery from the charger and place it in the charged battery box.
 
 If at any point throughout the charging cycle the charger indicates an issue (e.g. a timeout, capacity or cable error), disconnect the battery from the charger, apply a piece of red tape to the yellow XT60 connector on the battery and place it in the damaged battery box. Inevitably there will be a small number of faulty chargers. If you suspect that a particular charger is faulty (lots of batteries fail to charge on it), disconnect it from its power supply, apply red tape to its yellow XT60 connector and place it in the damaged battery box.
--- a/docs/competition/team-support/battery-charging.md
+++ b/docs/competition/team-support/battery-charging.md
@@ -10,8 +10,8 @@ We (Student Robotics) manage the charging of all batteries at the competition. B
 ## Setup
 
 ### Equipment
-
-There are two locations involved in battery charging:
+Most batteries and chargers will arrive with teams, but any batteries, bags, and chargers in inventory should be prepared during setup.
+There are two locations involved in battery charging. The equipment required at each is as listed below:
 
 - Helpdesk
     - 1 x Discharged Battery Box (18L RUB), clearly labelled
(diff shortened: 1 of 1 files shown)


docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
 
 ## Safety
 
 The biggest risk of damage or injury is from the banana plugs coming unplugged from the charger. These are the red and black plugs that go into the side of the charger. All of the chargers should have the cables cable-tied to the charger to make it difficult to remove these plugs. Under no circumstances should these red and black plugs be unplugged from the chargers.
 
 The batteries themselves have many built in safety features and, unless severely abused, are completely safe. All batteries undergoing charging must be placed inside a fireproof charging bag. In the extremely unlikely event of a fire due to a battery DO NOT attempt to tackle the fire - the batteries contain a flammable metal (Lithium) which cannot be safely tackled with most fire extinguishers.
 
-If at any point you have any doubts about safety, immediately summon the Competition Team Support Coordinator (Rich Barlow) and the Health & Safety Coordinator (Scarzy) via [Helpdesk](/competition/team-support/helpdesk).
+If at any point you have any doubts about safety, immediately ask the helpdesk to summon the Health & Safety officer.
```
PeterJCLaw: Why is there an explicit layer of indirection here? If there's a safety issue, surely we should go straight to the H&S officer.

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
+### Safe Use of Battery Charging Equipment - Battery Charging Station 
+
+- There should be `0.5A` of mains electricity available per charger.
+- It should be ensured that any power distribution / 4 ways is setup properly and able to handle the current draw.
+- Additionally, there should be no trailing wires across the floor to prevent charging batteries falling on the floor.
 
 ## Volunteer Requirements
 
-Every volunteer performing the Battery Charging role must have read this document fully. Any doubts must be discussed with Rich Barlow ASAP.
+Every volunteer performing the Battery Charging role must have read this document fully. Any doubts must be raised before beginning the role.
```
PeterJCLaw: This implies that doubts which emerge after beginning are prohibited, which could lead to people ignoring a problem. Please could we keep the earlier wording (even if we change the person to contact).

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
+There are two locations involved in battery charging:
+
+- Helpdesk
+    - 1 x Discharged Battery Box (18L RUB), clearly labelled
+    - 1 x Charged Battery Box (18L RUB), clearly labelled
+    - 1 x LiPo Bag for damaged batteries, clearly labelled, red taped
+    - At least one roll of red insulating tape
+- Battery Charging Station
+    - At least `n` battery chargers
+    - Enough LiPo charging bags for every charger.
```
PeterJCLaw: I don't think these two requirements make sense -- the majority of our chargers and bags are with teams and are only collected when they arrive. As a result, the charging station cannot have them set up until the teams arrive.

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
 
-## Equipment
+## Setup
 
- * Battery chargers
- * Discharged battery box
- * Charged battery box
- * Damaged battery/charger box
- * Red insulating tape
- * Example good batteries (one brand new, one squidgy)
```
PeterJCLaw: Why are we removing these example good batteries from the requirements?

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
 
 ## Procedures
 
-### Swapping a team's battery
+### Swapping a team's battery - Helpdesk
+
+- Teams will bring their discharged battery to the Helpdesk
+- Discharged batteries should be placed in the discharged battery box
+    - If the battery is damaged, red tape should be placed on it and it should be placed in the damaged LiPo bag instead of the discharged battery box. 
+- A charged battery should be retrieved from the charged battery box and handed to the team.
```
PeterJCLaw: We seem to have lost the reasoning behind this. It's pretty important that the runbook records _why_ things are done the way that they are so that we know what's reasonable to change about the way it's done.

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
- 1. If the battery is damaged, tape over the yellow XT60 connector with red tape and place it in the damaged battery box.
- 1. If the battery is not damaged, plug the yellow XT60 connector into the charger cable and the white balance connector into the 4-pin socket on the side of the charger.
- 1. Place the battery in the charging bag next to the charger.
- 1. Hold down the Start/Enter button for about 3 seconds until the screen displays `R:3SER S:3SER CONFIRM(ENTER)`. If both numbers are not 3 then disconnect the battery and summon Rich Barlow via Helpdesk.
- 1. If the display is as expected, press Start/Enter to commence charging.
+- Check that the battery is not physically damaged. There should be no splits or tears in the body of the battery, no nicks in the wire insulation, no exposed wire and the battery should not be so swollen that it is no longer squidgy. There will be a couple of example batteries at the Battery Charging station to compare to.
```
PeterJCLaw: This is an ordered list (and the ordering is an important part of the documentation), why are we removing the numbering here?

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
+
+- Charged LiPos should be taken to the helpdesk.
+- Discharged and damaged LiPos should be taken to the battery charging station.
+- Any batteries that were marked as damaged by the helpdesk, should be inspected and placed in the damaged battery box if they are damaged.
 
-Teams will bring discharged batteries to the Battery Charging station and require a replacement charged battery. A Battery Charging volunteer should receive the discharged battery from the team and place it in the discharged battery box. A charged battery should be removed from the charged battery box and handed to the team. Do not allow teams to take batteries from the box themselves - we want to prevent teams from hoarding batteries.
+### Charging a discharged battery - Battery Charging Station
 
-### Charging a discharged battery
+All of the battery chargers will be set-up and ready to use. Before a battery is connected the screen must display `LiPo BALANCE 2.2A 11.1V(3S)`. If it displays anything else then please do not use it and ask Helpdesk to summon help.
```
PeterJCLaw: Who (which role) should helpdesk summon at this point?

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
- 1. Place the battery in the charging bag next to the charger.
- 1. Hold down the Start/Enter button for about 3 seconds until the screen displays `R:3SER S:3SER CONFIRM(ENTER)`. If both numbers are not 3 then disconnect the battery and summon Rich Barlow via Helpdesk.
- 1. If the display is as expected, press Start/Enter to commence charging.
+- Check that the battery is not physically damaged. There should be no splits or tears in the body of the battery, no nicks in the wire insulation, no exposed wire and the battery should not be so swollen that it is no longer squidgy. There will be a couple of example batteries at the Battery Charging station to compare to.
+- If the battery is damaged, tape over the yellow XT60 connector with red tape and place it in the damaged battery box.
+- If the battery is not damaged, plug the yellow XT60 connector into the charger cable and the white balance connector into the 4-pin socket on the side of the charger.
+- Place the battery in the charging bag next to the charger.
+- Follow the instructions as found in the [Student Robotics kit docs](https://www.studentrobotics.org/docs/kit/batteries/).
```
PeterJCLaw: I believe in the past we have tried to use only a single type of charger on the battery charging station, in order to simplify this safety-important role.

Whether or not we diverge from that, I think it's important that the document which is present on the charging station (whether that be an exact printout of this page or something built from it) be a complete document rather than one which links to external sources. We should not assume that volunteers have any access to other resources.

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
+### Swapping batteries - Shuffling
+
+The volunteer that is running the battery charging station will periodically have to go and get discharge batteries from the helpdesk, simultaneously dropping off charged batteries. The helpdesk may also contact the volunteer to do it sooner.
+
+- Charged LiPos should be taken to the helpdesk.
+- Discharged and damaged LiPos should be taken to the battery charging station.
+- Any batteries that were marked as damaged by the helpdesk, should be inspected and placed in the damaged battery box if they are damaged.
 
-Teams will bring discharged batteries to the Battery Charging station and require a replacement charged battery. A Battery Charging volunteer should receive the discharged battery from the team and place it in the discharged battery box. A charged battery should be removed from the charged battery box and handed to the team. Do not allow teams to take batteries from the box themselves - we want to prevent teams from hoarding batteries.
+### Charging a discharged battery - Battery Charging Station
```
PeterJCLaw: nit: there's trailing whitespace on this line, please remove it.

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,76 @@ original:
...
+If at any point you have any doubts about safety, immediately ask the helpdesk to summon the Health & Safety officer.
 
 ## Procedures
 
-### Swapping a team's battery
+### Swapping a team's battery - Helpdesk
+
+- Teams will bring their discharged battery to the Helpdesk
+- Discharged batteries should be placed in the discharged battery box
+    - If the battery is damaged, red tape should be placed on it and it should be placed in the damaged LiPo bag instead of the discharged battery box.
```
PeterJCLaw: This means that volunteers on _both_ helpdesk and battery charging (which may be in different locations) need training in what damaged batteries look/feel like. What benefit does this serve over letting that be done only at the battery charging station?

docs/competition/team-support/battery-charging.md
jcwilliamson: This was an oversight on my part when making these changes. I will add them to the equipment for charging station.

docs/competition/team-support/battery-charging.md
jcwilliamson: I will remove these for the next PR. Is this list otherwise fine?

docs/competition/team-support/battery-charging.md
jcwilliamson: I plan to change this to a role title, rather than having a name. Is the team support co-ordinator still a role, or should this be replaced with a different role name?

PeterJCLaw: I'm not sure, though I'm not sure that's necessarily the right person anyway. It needs to be whoever is doing the training for this role, though I don't think we have a specific name for that.

docs/competition/team-support/battery-charging.md
PeterJCLaw: I think so. Maybe note here explicitly that the chargers and bags will arrive when the teams arrive, though we should try to set up as many as we can from our other stock?

For this year (SR2019), I think we'll likely be able to fully set up ahead of time as we have many more chargers than teams, though I don't expect that to be the usual case.

docs/competition/team-support/battery-charging.md
jcwilliamson: Should I possibly just direct this to health and safety aswell?

docs/competition/team-support/battery-charging.md
```diff
@@ -7,44 +7,74 @@ original:
...
+### Safe Use of Battery Charging Equipment - Battery Charging Station 
+
+- There should be `0.5A` of mains electricity available per charger.
+- It should be ensured that any power distribution / 4 ways is setup properly and able to handle the current draw.
+- Additionally, there should be no trailing wires across the floor to prevent charging batteries falling on the floor.
 
 ## Volunteer Requirements
 
-Every volunteer performing the Battery Charging role must have read this document fully. Any doubts must be discussed with Rich Barlow ASAP.
+Every volunteer performing the Battery Charging role must have read this document fully. Any doubts must be raisedwith the Health and Safety officer ASAP.
```
PeterJCLaw:
```suggestion
Every volunteer performing the Battery Charging role must have read this document fully. Any doubts must be raised with the Health and Safety officer ASAP.
```

docs/competition/team-support/battery-charging.md
PeterJCLaw: My interpretation of this paragraph is that it's for things which the volunteer is unsure about in regards to how they should perform the role. I'm not sure that the Health and Safety officer is a good person for them to ask questions of here -- the person who trained them is a much better person to ask.

docs/competition/team-support/battery-charging.md
RealOrangeOne: Battery charging will either have a radio, or be very close to one, which can be used to signal the H&S coordinator.

docs/competition/team-support/battery-charging.md
jcwilliamson: What reasoning do you think is needed here? Isn't this the way that battery charging is normally conducted.

docs/competition/team-support/battery-charging.md
jcwilliamson: Which role should be summoned here? There was nothing previously written about which role should be summoned.

docs/competition/team-support/battery-charging.md
jcwilliamson: Wouldn't it be wise for both sites to be checking for battery damage? I feel that it's better to be safe than sorry in this case.

docs/competition/team-support/battery-charging.md
PeterJCLaw: Again I think this should go to the person who is overall responsible for the battery charging and likely the person who did the training for the role.

Apologies, my question here was meant to be identifying that we should specify who to ask for help -- we need to ensure that there aren't any dead ends in the instructions we're providing to ensure that things are clear on the day.

docs/competition/team-support/battery-charging.md
PeterJCLaw: This is indeed the way it's normally conducted, however the runbook is about more than just recording _what_ is done. It's important to record _why_ things are done a particular way too. 

Without an understanding of why a piece of documentation suggests a given thing, it can be easy to end up following the docs "because the docs say so" and then either getting stuck unable to change a process, or ending up changing a process and finding out the hard way that things were the way they were for a good reason. Recording the reason upfront avoids both.

Recording the "why" of things is a general principle we're trying to keep in the runbook, though as it's based on lots of existing docs not everything has that yet. It feels like it would be a shame for this particular bit to go backwards when it already has the reasoning recorded.

docs/competition/team-support/battery-charging.md
PeterJCLaw: Possibly, however the overhead in training is non-trivial. There is also the possible outcome that each assumes the other has done a good enough check and then doesn't check again.

In my experience shared responsibilities are ok if all parties are co-located and constantly working closely with each other, but otherwise result in everyone thinking that the job anyone could do would be done by someone else. I'd therefore encourage we make this a specific responsibility of the battery-charging station.

I'd be happy for @Scarzy to overrule me here though if he wanted.

PeterJCLaw:
@jcwilliamson apologies for the delayed response here, I've had another look and hopefully clarified some of my suggestions.

sedders123:
I'm aware @jcwilliamson is unable to take on any tasks currently. I'm happy to make these changes on Saturday at the Hackday if needed.

docs/competition/team-support/battery-charging.md
Scarzy: > I'd therefore encourage we make this a specific responsibility of the battery-charging station.

I'd agree with this. It is the responsibility of battery charging to do the check., otherwise we run the risk of some being overlooked.

However, that does not mean that helpdesk shouldn't apply common sense, if a battery turns up emitting smoke I would be rather disappointed if they just put it in the normal box and ignored it...

I'm fairly sure in previous years that the batteries have _always_ been stored in the fireproof bags. Helpdesk put them in a bag when received from a team, transport them to battery charging in said bag, and after charging the batteries get returned to and stored at helpdesk in a bag. In which case it becomes a much smaller risk, and a bit moot for helpdesk to do any checking anyway - the handling procedure would be the same.

sedders123:
@jcwilliamson Are you still working on this, or would it be useful for someone else to take over the last review changes?

jcwilliamson:
If someone else could take over that would be good, thanks.

Adimote:
I'm going to close this, as I took on the work and incorporated it in a new PR, it seems there aren't any outstanding comments.

PR #29, opened 2019-03-31
Update reception flowchart

PeterJCLaw:
Updates the age check to 16 and fixes a bug where underage competitors could sign their own MCFs.

image


PeterJCLaw:
Yeah, I don't think the diff is going to be useful here -- it was previously effectively SVG-as-output-blob from OmniGraffle.

PR #30, opened 2019-03-31
Add procedure for handling an incident occuring

Scarzy:
This fixes https://github.com/srobo/tasks/issues/254

### DIFF
--- a/docs/competition/event/incident-management.md
+++ b/docs/competition/event/incident-management.md
@@ -46,6 +46,7 @@ Due to the impact of stopping the competition, ability to do so is restricted to
 * Majority decision from Trustees.
 
 Stopping the competition will be indicated by a call over the radio of "This is _role_; stop the competition."
+This call should be repeated at least once, and every minute or so until acknowledged by one of the other aforementioned roles.
 
 At this point the following should occur;
 
--- a/docs/competition/event/incident-management.md
+++ b/docs/competition/event/incident-management.md
@@ -1,6 +1,6 @@
 # SR2019 Incident Management Procedure
 
-This page is primarily for internal use. If you are a team competing in Student Robotics, you probably want our main website.
+This page is primarily for internal use. If you are a team competing in Student Robotics, you probably want [our main website](https://studentrobotics.org/).
 
 This page details the procedures to be following in the event of an incident occuring during the competition. Incidents are split into 2 types; minor and major.
 
--- a/docs/competition/event/incident-management.md
+++ b/docs/competition/event/incident-management.md
@@ -1,0 +1,85 @@
+# SR2019 Incident Management Procedure
+
+This page is primarily for internal use. If you are a team competing in Student Robotics, you probably want our main website.
+
+This page details the procedures to be following in the event of an incident occuring during the competition. Incidents are split into 2 types; minor and major.
+
+What constitutes an incident is rather unpredictable, and as such is not fully defined. Common sense should prevail in determining the nature and severity of the incident.
+
+This procedure is superceded by any requirements of the venue, e.g. in the event of a fire follow the venues' fire evacuation procedure.
+
+# Minor Incidents
+
+A minor incident is anything that needs some kind of response but has minimal impact on the event.
+
+Examples may be;
+
+* A competitor has twisted their ankle on a stairwell,
+* Infrastructure cabling has become damaged and poses a hazard.
+
+Procedure here is simple;
+
+* Notify relevant coordinator,
+* Follow their instructions
+
+# Major Incidents
+
+A major incident is anything that causes a significant impact on the event as a whole.
+
+Examples might include;
+
+* Accident requiring significant first aid,
+* Battery failure renders access to an area unsafe
+
+As with minor incidents, the initial reaction should be to notify the relevant coordinator and follow their instructions. If first aid is required the First Aid Provider should __always__ be notified first.
+
+## Stopping the competition
+
+A major inicident may call for the competition to be stopped. This may be at the end of a current match or in the middle of a match.
+
+Due to the impact of stopping the competition, ability to do so is restricted to the following roles:
+
+* Health and Safety Cooordinator,
+* Event Coordinator,
+* Head Judge,
+* Majority decision from Trustees.
+
+Stopping the competition will be indicated by a call over the radio of "This is _role_; close the arena."
+
+At this point the following should occur;
+
+* Matches stop,
+* Power tools area to be vacated,
+* Battery charging to be turned off,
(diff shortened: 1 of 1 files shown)


docs/competition/event/incident-management.md
```diff
@@ -0,0 +1,85 @@
...
+
+* Matches stop,
+* Power tools area to be vacated,
+* Battery charging to be turned off,
+* Competitors, Team Leaders, and members of the public should be asked to remain in their current location,
+* Competitors to be asked to stop testing on robots.
+
+The aim of these is to prevent any other issues occuring whilst the current incident is being dealt with, and to prevent attendees from getting in the way of the handling the incident.
+
+## Shift in roles
```
Adimote: Should we document this information in each of those roles?

docs/competition/event/incident-management.md
```diff
@@ -0,0 +1,85 @@
+# SR2019 Incident Management Procedure
+
+This page is primarily for internal use. If you are a team competing in Student Robotics, you probably want our main website.
```
Adimote:
```suggestion
This page is primarily for internal use. If you are a team competing in Student Robotics, you probably want [our main website](https://studentrobotics.org/).
```

docs/competition/event/incident-management.md
```diff
@@ -0,0 +1,85 @@
...
+A major inicident may call for the competition to be stopped. This may be at the end of a current match or in the middle of a match.
+
+Due to the impact of stopping the competition, ability to do so is restricted to the following roles:
+
+* Health and Safety Cooordinator,
+* Event Coordinator,
+* Head Judge,
+* Majority decision from Trustees.
+
+Stopping the competition will be indicated by a call over the radio of "This is _role_; close the arena."
```
Adimote: I think we need a different magic phrase here, we use the term "close the arena" too much when we close it after tinker-time. Maybe "Halt the competition"?

docs/competition/event/incident-management.md
Scarzy: Potentially, though really all volunteers need to be familiar with this particular process, as if enacted it impacts everyone.

docs/competition/event/incident-management.md
Scarzy: Perhaps. I might wait and see if any other suggestions come out of the woodwork.
Whatever it is, it needs to be something that doesn't immediately make anyone overhearing it panic, while being sufficiently clear as to the intent.

RealOrangeOne: Maybe it's not worth defining the exact message to be used. If we're defining this as 'stopping the competition', then simply saying 'stop the competition' would work, likely with a 'I repeat...' part to ensure it's heard correctly.

Alternatively, and in no way seriously, code words with specific intent also work, like say [pineapple](https://youtu.be/gcL8K0QqhJA?t=31).

docs/competition/event/incident-management.md
```diff
@@ -0,0 +1,85 @@
...
+# Major Incidents
+
+A major incident is anything that causes a significant impact on the event as a whole.
+
+Examples might include;
+
+* Accident requiring significant first aid,
+* Battery failure renders access to an area unsafe
+
+As with minor incidents, the initial reaction should be to notify the relevant coordinator and follow their instructions. If first aid is required the First Aid Provider should __always__ be notified first.
```
PeterJCLaw: nit: maybe put the first aid part first, so that anyone reading this while responding to an incident (hopefully won't happen though I suspect people's order of consumption of information reflects how they use it) are prompted to seek first aid first?

```suggestion
If first aid is required the First Aid Provider should __always__ be notified first.
Following that, as with minor incidents, the reaction should be to notify the relevant coordinator and follow their instructions. 
```

docs/competition/event/incident-management.md
```diff
@@ -0,0 +1,85 @@
...
+* Event Coordinator,
+* Head Judge,
+* Majority decision from Trustees.
+
+Stopping the competition will be indicated by a call over the radio of "This is _role_; close the arena."
+
+At this point the following should occur;
+
+* Matches stop,
+* Power tools area to be vacated,
```
PeterJCLaw: Should we be turning off the power tools first? If so, maybe state that?

PeterJCLaw: (Ah, just spotted this is mentioned below)

docs/competition/event/incident-management.md
```diff
@@ -0,0 +1,85 @@
...
+
+If the match is stopped immediately, if safe to do so, the Marshalls should hit the power cut off on each robot, and then vacate the arena.
+
+In either scenario; the Badgers should ask all competitors to leave the staging area. Robots to remain where they are, power cut offs to be hit.
+
+## Resuming/cancelling the competition
+
+Resuming the competition requires the consent of all involved in stopping it.
+
+Cancelling the event requires a majority vote from Trustees.
```
PeterJCLaw: Should there be a fallback here in case the trustees are unavailable?

I _think_ all the trustees will be at this year's event, but I'm not sure that will always be the case.

Scarzy: Probably. Perhaps it should be limited to trustees that are present, and extended to include the roles mentioned above?

docs/competition/event/incident-management.md
Scarzy: Thinking about it, any significantly affected roles should probably just have a comment saying to check their responsibilities under this. That way we have a single source of truth should an incident ever arise, rather than any confusion from this differing from the other documents.

docs/competition/event/incident-management.md
Scarzy: I didn't want competitors hovering around power tools longer than necessary, hence why it's down to the roving helper to shut it all down.

docs/competition/event/incident-management.md
Scarzy: We want the wording to be reasonably specific to avoid another message being misinterpreted as stopping the competition.

I've changed the wording, and required repeating it.

docs/competition/event/incident-management.md
PeterJCLaw: Yeah, or maybe trustees first and then, if there aren't a quorate number present then encourage them to consult with the above roles?

PeterJCLaw:
@Scarzy ping on the status of this? I don't feel strongly about the cancellation logic, I'd hope that good sense would prevail in the event the trustees are unreachable; do you want to make a change to that, or can we merge this as-is?

Scarzy:
Sorry for delay.

I think good sense has to prevail, otherwise we'll constantly have a "but
what if...?" issue. So I'd say merge as-is.

PR #34, opened 2019-04-22
Core volunteering docs

PeterJCLaw:
This adds some basic getting started guidance and provides a root page for our now extant volunteering section. It also introduces how tasks work and a glossary of SR specific terms.

This draws some ideas from https://www.studentrobotics.org/cgit/volunteer-handbook.git/ (as well as external sources), though deliberately doesn't attempt to import it because much of the content there doesn't fit into the core (and some is out of date). Likely we can use that as a reference for other pieces later.


docs/volunteering/tasks.md
```diff
@@ -0,0 +1,43 @@
+# Tasks
+
+We record tasks which need doing at https://github.com/srobo/tasks/issues.
+
+If you spot a task which you'd like to do, please `@mention` the person who
+created the task to let them know that you're interested. If you're able to
+self-assign the task (for technical reasons you won't initially be able to do
+this), please also do that.
+
+<!--- Review question: is the original author the right person to ping here? --->
```
PeterJCLaw: (obviously this line will be removed before merging)

Thoughts on this?

My reasoning is that the author is likely able to pass the volunteer on to someone else.
I am however a little concerned that this might put people off opening issues if they feel they'll end up needing to own the issue somewhat.

docs/volunteering/index.md
```diff
@@ -0,0 +1,33 @@
...
+for contact details and for you to read the [Code of Conduct][code-of-conduct].
+
+They'll also be able to answer any questions you have about the organisation and
+get you set up with accounts on our collaboration platforms (we primarily use
+[Slack][sr-slack] and [GitHub][sr-github]).
+
+[code-of-conduct]: https://srobo.gitbook.io/ops-manual/about-the-charity/code-of-conduct
+[sr-slack]: https://studentrobotics.slack.com/
+[sr-github]: https://github.com/srobo
+
```
PeterJCLaw: I'm probably going to introduce a "What we expect from you" section here, but I'm not sure if that'll make it into this PR. I think we need such a section, but clearly the wording needs to be both clear and empathetic and therefore needs care in its crafting.

docs/volunteering/tasks.md
```diff
@@ -0,0 +1,55 @@
...
+    Please do _not_ assign other people to tasks unless they have specifically
+    said that they are doing the task.
+
+If you're asked whether you can take on a task, there is no obligation to do so.
+However please do ensure that you respond in a timely fashion so it's clear you
+won't be able to take on a task.
+
+If you've taken on a task and find that you no longer have the time or energy to
+complete it, that's fine! However please do post on the task mentioning that
+you're no longer able to complete it and remove your assignment from the task.
```
trickeydan: NB: Most volunteers cannot currently unassign themselves.

docs/volunteering/glossary.md
```diff
@@ -0,0 +1,66 @@
...
+
+: this is the canonical way to refer to a given year's competition programme;
+the year number is the calendar year in which the competition event takes place,
+so SR2019 was the _competition programme_ which ran from November 2018 until
+April 2019
+
+_Tech Day_
+
+: during the _competition year_ we organise events at which the teams can come
+with their robots and get extra help from _mentors_ as well as meet other teams
```
trickeydan: Some glossary suggestions:

- TLA
- SRComp
- Rookie Team
- Token, and how this differentiates from a Marker

docs/volunteering/tasks.md
RealOrangeOne: This is false. Anyone who can be assigned an issue also has the permission to unassign themselves, and change anyone else who's assigned.

(At time of writing, this applies to *everyone* with access to the repo, either through the org or as a collaborator. If it happens in future, it'll be because they've been added since and were incorrectly setup)
(If i'm wrong, send me a username of someone who has the issue now and i'll look into it)

docs/volunteering/tasks.md
RealOrangeOne: It's not ideal, but i doubt there's an alternative. Having the person open them tag the relevant coordinator might work, but some things will transcend. 

Would relying on the fact that those who will likely be tagged are also likely watching the repo be enough? I somewhat doubt so? Alternatively someone watching can point in the right direction? eg:

> Person 1: What needs to be done to do this?
Person 2: @Person_3 will know how to do that
Person 3: Yes, do ...

It feels a little bureaucratic, but would probably work well. Obviously if 'Person 2' can answer the question, they should.

docs/volunteering/tasks.md
trickeydan: If the text reflects the intended state, then that's fine. I'll have a chat with the people who've mentioned this to me. Even if it is still the case, this is not the place to discuss it.

docs/volunteering/index.md
trickeydan: "What we expect from you" is definitely not the right wording. How about "Volunteer responsibilities"?

docs/volunteering/glossary.md
PeterJCLaw: Hrm.
I'm not sure there's a benefit in this glossary merely encompassing all possible terms (i.e: a dictionary), nor do I want it to become either a duplicate of the rulebook or other resources.

For example: I'd hope that "SRComp" would be sufficiently googleable and "Rookie Team" clear enough from its constituent words' meaning.

Will have a think about how to draw the line here.

PeterJCLaw: Turns out: "SRComp" is semi-googleable. You do get some results which I'd call "right", but many which aren't.

PeterJCLaw: 4b3e20a

docs/volunteering/index.md
PeterJCLaw: c18737cec636f55457ed2fb0942d444122a1dee1 adds such a section.

docs/volunteering/tasks.md
PeterJCLaw: 38c4d486c6b4af86e14368935e76e20afb8e6bb2 expands on this a bit. I'm not completely happy with it, but I'm not sure of a better suggestion to make or a better way to frame it.

Suggestions for improvements are welcome, though I suggest that unless there's anything major we go with this for a while and if a better solution appears fix it then.

docs/volunteering/tasks.md
```diff
@@ -0,0 +1,61 @@
+# Tasks
```
RealOrangeOne: Would it be worth as part of this page talking about milestones and how we use them?

PeterJCLaw: Possibly, if we you think we do anything here which is different to how they're used in general?

Unless you think that's a blocker I'd also like to get this merged and iterate from there, rather than continue to develop here.

docs/volunteering/glossary.md
```diff
@@ -0,0 +1,70 @@
...
+: a volunteer at the competition event who helps ensure teams make it to their
+matches on time
+
+[_Slack_](https://studentrobotics.slack.com/):
+
+: an instant-messaging tool we use for quick discussions
+
+_SR_:
+
+: a shortened form of Student Robotics
```
Adimote: Just making this a bit clearer that it's a term, not a shortened version of the competition itself

```suggestion
: a shortening of 'Student Robotics'
```

docs/volunteering/glossary.md
```diff
@@ -0,0 +1,70 @@
...
+
+[_SRComp_](https://github.com/PeterJCLaw/srcomp/wiki):
+
+: a software stack which we use to run competition events
+
+_srobo_:
+
+: another shortened form of Student Robotics
```
Adimote: Same as above, plus a mention of SR above
```suggestion
: another shortening of 'Student Robotics' (same as _SR_)
```

docs/volunteering/tasks.md
```diff
@@ -0,0 +1,61 @@
...
+
+We record tasks which need doing at https://github.com/srobo/tasks/issues.
+
+If you spot a task which you'd like to do or help with, please `@mention`
```
Adimote: is it worth linking to https://help.github.com/en/articles/basic-writing-and-formatting-syntax#mentioning-people-and-teams ? I'm not sure how common @mentions are outside of github and slack.

(And the 2 times below too)

docs/volunteering/glossary.md
PeterJCLaw: 40d6595f7a18790e96b6ad8ec0d4566e96abbdb4

docs/volunteering/glossary.md
PeterJCLaw: 40d6595f7a18790e96b6ad8ec0d4566e96abbdb4

docs/volunteering/tasks.md
PeterJCLaw: `@mention`s worked on IRC and I'm fairly sure they work on Twitter?
That's a useful link to add though: ce9b8b9

docs/volunteering/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Being empathetic and creating a safe environment
+
+We should be thinking about others whenever we do something, acknowledging the
+effort they may have put in before criticising.
+
+For example: if you're reviewing something a colleague has been working hard on,
+and all you do is say "Approved, but needs a couple of changes" then you haven't
+thought about the effect that has on the person, nor have you considered the
+power of saying something more congratulatory and celebratory.
```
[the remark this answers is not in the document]
PeterJCLaw: Ta. I may have borrowed this heavily from a similar document at work.

docs/volunteering/glossary.md
```diff
@@ -0,0 +1,70 @@
...
+_Competition programme_:
+
+: the overall programme of events and activities which the teams participate in
+
+_Competition year_:
+
+: the 6 months (or so) between _Kickstart_ and the _Competition_ during which
+teams build their robots; see also _SRnnnn_
+
+_Core Team_:
```
trickeydan: This is outdated. I suggest replacing this with definitions of the Competition and Kit Team.

docs/volunteering/glossary.md
```diff
@@ -0,0 +1,70 @@
...
+
+: the group of 8 or so volunteers responsible for delivering a particular year's
+_Competition programme_.
+
+[_GitHub_](https://github.com/srobo):
+
+: a platform for hosting code repositories which we also use to coordinate tasks
```
trickeydan: I personally think it would be better to focus on GitHub as a project management tool rather than as a code hosting platform. Code can be scary for new volunteers, and ultimately we do use it primarily for organisation.

docs/volunteering/glossary.md
RealOrangeOne: I don't think it should be replaced, but I think saying it's what the Competition Team was referred to as for the 2019 competition cycle is definitely worthwhile.

docs/volunteering/glossary.md
RealOrangeOne: This definition is accurate for what GitHub is, and what we use it for. 

> GitHub is a development platform

src: https://github.com/

Perhaps something like:

```suggestion
: a development platform, used to host code repositories and to coordinate tasks
```

docs/volunteering/glossary.md
trickeydan: At the very least, the definitions of the competition and kit team should be here.

PeterJCLaw: a13a57b

docs/volunteering/glossary.md
PeterJCLaw: 5f141692c693428140e4fdb47fa11a6df0c26db7

PR #55, opened 2019-08-26
Add meta information

### LINKED ISSUE #23: Tidy up the root page
Currently the root page is empty. We should probably use that space for an overview page, including guidance on roughly what's where or how to get stuck in.

PeterJCLaw:
This fills out the root page with meta information about the runbook as well as including a brief summary in the root readme.
I'm not actually sure if the (hosted) homepage is the right place for this, but it feels like a useful place to start.

Fixes #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ and [Kit Team][kit-team], as well as other activities.
 
 ## Installation
 
-In order to bootstrap this project you'll need:
+To run this project locally, you'll need:
 
 - Python (`>=3.5`)
 - `pip`
--- a/README.md
+++ b/README.md
@@ -1,12 +1,10 @@
-# RunBook
-
-Student Robotics Competition Program
+# Student Robotics Volunteer RunBook
 
 [![CircleCI](https://circleci.com/gh/srobo/runbook.svg?style=svg)](https://circleci.com/gh/srobo/runbook)
 
-# Installation
+## Installation
 
-## Requirements
+In order to bootstrap this project you'll need:
 
 - Python (`>=3.5`)
 - `pip`
--- a/README.md
+++ b/README.md
@@ -2,6 +2,13 @@
 
 [![CircleCI](https://circleci.com/gh/srobo/runbook.svg?style=svg)](https://circleci.com/gh/srobo/runbook)
 
+The runbook aims to be a single source for all Student Robotics internal
+documentation. This includes information for the [Competition Team][competition-team]
+and [Kit Team][kit-team], as well as other activities.
+
+[competition-team]: https://opsmanual.studentrobotics.org/annual-robotics-competition/competition-programme-team
+[kit-team]: https://opsmanual.studentrobotics.org/v/master/annual-robotics-competition/kit-team
+
 ## Installation
 
 In order to bootstrap this project you'll need:
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,9 +4,6 @@ Welcome to our runbook. This aims to be a single source for all Student Robotics
 internal documentation, including information for the [Competition Team][competition-team]
 and [Kit Team][kit-team] as well as other activities.
 
-[competition-team]: https://opsmanual.studentrobotics.org/annual-robotics-competition/competition-programme-team
-[kit-team]: https://opsmanual.studentrobotics.org/v/master/annual-robotics-competition/kit-team
-
 ## Content Guidance
 
 The runbook aims to document all the Student Robotics specific information which
@@ -33,4 +30,6 @@ There are many cases where it should not contain the actual documentation:
    tooling, or recipes for using tooling which are common to us but too specific
    for general documentation.
 
+[competition-team]: https://opsmanual.studentrobotics.org/annual-robotics-competition/competition-programme-team
+[kit-team]: https://opsmanual.studentrobotics.org/v/master/annual-robotics-competition/kit-team
 [ops-manual]: https://opsmanual.studentrobotics.org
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -1,12 +1,17 @@
...
-
-Student Robotics Competition Program
+# Student Robotics Volunteer RunBook
 
 [![CircleCI](https://circleci.com/gh/srobo/runbook.svg?style=svg)](https://circleci.com/gh/srobo/runbook)
 
-# Installation
+The runbook aims to be a single source for all Student Robotics internal
+documentation. This includes information for the [Competition Team][competition-team]
+and [Kit Team][kit-team], as well as other activities.
```
trickeydan: The Kit Team Committee are due to discuss where we want to document things in our first meeting.

I agree it would make sense to put it here.

README.md
```diff
@@ -1,12 +1,17 @@
...
+
+[competition-team]: https://opsmanual.studentrobotics.org/annual-robotics-competition/competition-programme-team
+[kit-team]: https://opsmanual.studentrobotics.org/v/master/annual-robotics-competition/kit-team
```
trickeydan: Worth noting that this is not the final revision and it is probably ill-advised to link to draft versions (`v/master`) of the Ops manual.

docs/README.md
```diff
@@ -0,0 +1,36 @@
+# Meta
+
+Welcome to our runbook. This aims to be a single source for all Student Robotics
+internal documentation, including information for the [Competition Team][competition-team]
+and [Kit Team][kit-team] as well as other activities.
+
+[competition-team]: https://opsmanual.studentrobotics.org/annual-robotics-competition/competition-programme-team
+[kit-team]: https://opsmanual.studentrobotics.org/v/master/annual-robotics-competition/kit-team
```
trickeydan: Same as https://github.com/srobo/runbook/pull/55/files#r317641799

README.md
trickeydan: Having read the other README file, I am now more unsure about where to document things. The runbook is a place for processes etc, which makes a lot of sense.

Open to suggestions about where we put technical documentation, but I doubt a decision will be made before the full Kit Team is formed.

README.md
PeterJCLaw: Yeah, I'm not sure how best to otherwise link to it though? Is there a way to link to a specific revision (git hash, not published version)?

PeterJCLaw:
> If the changes here are made, please can we update the description of the kit repo, as this would make it no longer specific to the Competition Programme.

I'm not sure what you mean by "kit repo" here?

Do you mean the https://github.com/srobo/kit-maintenance-docs/? If so, I'm expecting that those would stay as they are -- they're an example of tool-specific internal documentation which these docs would link to. (Maybe I should change that to "task specific"?)

(For clarity: what I've written up here has always been my intent for this documentation -- that it's a single source of general information)

README.md
PeterJCLaw: (An alternative might be to just not link to the kit team for now)

trickeydan:
> I'm not sure what you mean by "kit repo" here?

Complete typo on my behalf I'm afraid. I meant this repo (https://github.com/srobo/runbook/), where the description currently talks about the competition programme. This PR defines that scope as slightly wider than that, covering most processes within SR, notably the kit team, and presumably other teams if / when they form.

README.md
PeterJCLaw: Personally I think I'd expect that detailed technical documentation for the kits would lie within their own source repos where that's possible. If we do _need_ to have that sort of thing for, say, the ODROIDs I think it would be fine for it to be in here (though I'd hope we could work with the upstream in most cases).

PeterJCLaw:
> > I'm not sure what you mean by "kit repo" here?
> 
> Complete typo on my behalf I'm afraid. I meant this repo (https://github.com/srobo/runbook/), where the description currently talks about the competition programme. This PR defines that scope as slightly wider than that, covering most processes within SR, notably the kit team in addition to the competition team, and presumably other teams if / when they form.
> 
> I think that this is a good idea, I just wanted to make sure that the descriptions get appropriately updated.

Ah, yeah, good spot. I've never really understood how we ended up with that name there (I think it was a misunderstanding on the original task).

README.md
trickeydan: It won't be finalised until at least Wednesday, and that would be if the kit team committee are happy with it in it's current state.

I guess it depends on how soon you want this PR merged.

README.md
```diff
@@ -1,12 +1,17 @@
...
+
+## Installation
 
-## Requirements
+In order to bootstrap this project you'll need:
```
RealOrangeOne: Using _bootstrap_ in this context is a little strange to those who haven't seen it before, and not obvious what it means.

```suggestion
To run this project locally, you'll need:
```

docs/README.md
```diff
@@ -0,0 +1,36 @@
+# Meta
```
RealOrangeOne: I wonder if this title is necessary? It looks a little strange in the rendered output

image

PeterJCLaw: The default (which appears if you don't have a top level heading) is "Home". I'll change to that.

PeterJCLaw: 9e8a42a

README.md
PeterJCLaw: I was avoiding that phrasing as it's misleading -- not only are they not all that you need, they're also not all strictly needed anyway; instead they're the tools which let you install the actual dependencies.

PeterJCLaw: da1386d

docs/README.md
```diff
@@ -0,0 +1,36 @@
+# Home
+
+Welcome to our runbook. This aims to be a single source for all Student Robotics
+internal documentation, including information for the [Competition Team][competition-team]
+and [Kit Team][kit-team] as well as other activities.
+
+[competition-team]: https://opsmanual.studentrobotics.org/annual-robotics-competition/competition-programme-team
```
RealOrangeOne: I think it would be worth moving these links to the bottom of the file where it's standard to put links. Else inline with the content.

PeterJCLaw: I actually quite like having them close to their usages, but I don't feel strongly: 02f0f93

Issue #54, opened 2019-08-20
Import mentor guidance doc


PeterJCLaw:
I recently wrote a [mentor guidance doc](https://docs.google.com/document/d/1QEPMs5okJV5En-WPz5YgHt1_6kkx5C85YV-QP3oswQ0/edit) for volunteers at the Smallpeice summer school. It has general applicability, so it would be good to merge it into the runbook too.

PR #56, opened 2019-08-26
Import some guidance for mentors

PeterJCLaw:
This imports from a Google Doc some notes I wrote up before the Smallpeice 2019 summer school.

Fixes https://github.com/srobo/runbook/issues/54

I've just spotted that, after #34 merges, this might be better suited in the `/volunteering` section. I think it's fine to keep it here and then move it once that is in place and we have more of a feel for how that section will look.


docs/teams/mentor-guidance.md
```diff
@@ -0,0 +1,224 @@
...
+the trunk is a sapling then adding a large bough with many branches will
+crush the sapling. Conversely, presenting a bud of information can be
+frustrating for both parties if the person you’re helping is more
+experienced.
+
+For complete novices, or when very first introducing a new topic (even
+to an experienced audience), this means that you should try to present
+an idea in its very simplest form. Try to think about the essence of
+what makes the idea important, rather than the details of how it works.
+A good example of this is the idea of “[*lies to children*][lies-to-children]”.
```
trickeydan: Might be worth checking if `“` and `”` render correctly in mkdocs.

Alternatively: `"`

docs/teams/mentor-guidance.md
```diff
@@ -0,0 +1,224 @@
...
+[lies-to-children]: https://en.wikipedia.org/wiki/Lie-to-children
+
+As a concrete example, let’s consider the case of explaining variable
+assignment in Python:
+
+``` python
+foo = 42
+```
+
+A simple explanation here might be that we are “telling the computer to
```
trickeydan: In this section, it may be nicer to use **bold** than multiple varieties of quotes.

docs/teams/mentor-guidance.md
PeterJCLaw: They render fine :+1:. I'd kinda hoped that mkdocs would even convert plain ones, but it doesn't :disappointed:.

PeterJCLaw:
I've just spotted that, after #34 merges, this might be better suited in the `/volunteering` section. I think it's fine to keep it here and then move it once that is in place and we have more of a feel for how that section will look.

PeterJCLaw:
Actually, scratch that. The `/teams` section is aimed at the person organising teams. The `/volunteer` section is for volunteers. I'm going to move this and cross-link it.

docs/volunteering/mentor-guidance.md
```diff
@@ -0,0 +1,223 @@
...
+Once the learner has formed enough of a basis of understanding of a
+topic you can start asking them to explain how they think things might
+work if there were slight variations. In our Python example, you might
+show a conditional block:
+
+``` python
+foo = 42
+if foo > 25:
+    print('greater')
+print('nope')
```
RealOrangeOne: Are you missing an `else` here?

PeterJCLaw: probably, that's entirely not the point though.

PR #67, opened 2019-08-31
Switch to a pip-tools workflow

PeterJCLaw:
This is motivated by the issues around pipenv always upgrading every dependency whenever you touch any dependency.
(See https://github.com/pypa/pipenv/issues/2665 and https://github.com/pypa/pipenv/issues/2412)

This PR is partly to show what this would look like, as a point of comparison against the current setup.


trickeydan:
I'd personally rather we used a PEP 518 compliant tool, rather than yet another non-standard one.

.circleci/config.yml
```diff
@@ -8,28 +8,52 @@ jobs:
             - checkout
 
             - restore_cache:
-                key: cache-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
-            - run: pipenv install --deploy --dev
+                key: cache-{{ .Branch }}-{{ checksum "requirements.txt" }}
+            - run:
+                name: Install dependencies
+                command: |
+                  python3 -m venv venv
```
RealOrangeOne: AFAIK there's no need to run things inside a venv on CI, you can install direct to the main environment, which makes things look a little nicer.

scripts/build
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
RealOrangeOne: I've generally preferred `/usr/bin/env bash` for scripts like this, is there a benefit to hardcoding `/bin/sh`?

scripts/build
```diff
@@ -0,0 +1,3 @@
...
+
+exec mkdocs build
```
RealOrangeOne: Why are the commands prefixed with `exec`? They should work fine without it.

README.md
```diff
@@ -10,28 +10,27 @@ Student Robotics Competition Program
...
 ```
-pipenv install --dev
+pip install -r requirements.txt
 ```
 
 ### Running development server
 
 ```
-pipenv run serve
+mkdocs serve  # or ./scripts/serve
```
RealOrangeOne: As the commands are this short, I think it's ok to omit the scripts. The scripts are also not windows compatible (sans WSL), which it would be nice to maintain 1st-party support for.

Also, currently the scripts imply that the virtualenv has been activated, which isn't explicitly documented here, so we should probably document that!

.circleci/config.yml
PeterJCLaw: True. I'm too used to doing this anyway. It does have the advantage that it's most likely to be close to what contributors are doing.

scripts/build
PeterJCLaw: Mostly habit, though I'd definitely use `sh` over `bash`.

`sh` is asking for a POSIX compatible shell, which may or not happen to be `bash`. The former is _usually_ faster to start. Notably about five years ago Ubuntu (and maybe other Debian systems) sped up their boot time quite a lot by changing all their boot scripts to `#!/bin/sh` over `#!/bin/bash`.

`/bin/sh` is pretty close to guaranteed to be present in a Linux system. I'm not sure why that path should be any less likely to be present than `/usr/bin/env` (which is more common for launching other interpreters), so I don't find it compelling to launch two processes to run a script (`env` then the shell) over just the one which the current spelling does.

scripts/build
PeterJCLaw: To ensure that their exit codes are passed back to the originator. I _think_ `bash` will always do that, however it's a little bit magical and I'm not sure whether it's consistent between shells.

It also has the benefit of avoiding launching a fresh process.

README.md
PeterJCLaw: If there was tab completion on `mkdocs` I'd agree.

PeterJCLaw: I was actually wondering about including windows implementations for the scripts, so that if we happen to change the build setup again we could do so without needing to change what contributors need to run.

I've mixed feelings about documenting that the virtualenv needs to be active. We weren't doing do before and you don't seem to be advocating to do so for running the `mkdocs` commands directly. I'm therefore not sure why the scripts should be treated any differently.

In any of these cases (including the previous setup) the reader still needed to manage their environment themselves to ensure that the command being run was available. Nothing has changed in that regard.

RealOrangeOne: > you don't seem to be advocating to do so for running the mkdocs commands directly
Sorry, I was definitely suggesting we do this. They were separate suggestion points.

Including windows scripts could also work, but it's a lot of extra scripting. Having the scripts auto-source the virtualenv would definitely make them more useful in my eyes.

PeterJCLaw: > Having the scripts auto-source the virtualenv would definitely make them more useful in my eyes.

How do you propose that they do that? They won't know where the virtualenv is, nor even if one is needed. Setup of the virtualenv is up to the developer at the moment (this has not changed).

RealOrangeOne: Currently the scripts depend on `pipenv`, which handles the virtualenv stuff for the user. I think it feels fine that the new scripts also handle it. 

It does depend on the virtualenv being in a known location, which isn't especially ideal, but there are plenty of conventions around virtualenv naming, we just need to pick one.

trickeydan: This entire thread feels like we are reimplementing `pipenv` using an less standard tool. I'd personally rather we used a PEP 518 compliant tool, rather than yet another non-standard one.

> This is motivated by the issues around pipenv always upgrading every dependency whenever you touch any dependency.

Please could you give some specific examples of where this has caused an issue with *this repo*? I dislike `pipenv`, but it is probably the best tool for this project.

RealOrangeOne: https://studentrobotics.slack.com/archives/CEG05AAG6/p1567264146004600

trickeydan: > [studentrobotics.slack.com/archives/CEG05AAG6/p1567264146004600](https://studentrobotics.slack.com/archives/CEG05AAG6/p1567264146004600)

Okay, I agree that this is an issue.

However, I still don't think that we should be working around the limitations of `pip-tools` by hacking virtualenv activation into scripts. If that is a required feature of the toolset, I suggest we use a tool that does it. Alternatively, we could use pip-tools and provide instructions on how to use a virtualenv / `pip-tools`. A link to the appropriate docs is sufficient imo.

PeterJCLaw: > Currently the scripts depend on `pipenv`, which handles the virtualenv stuff for the user. I think it feels fine that the new scripts also handle it.

Well, sort-of. In reality `pipenv` isn't going to be installed for most people by default, either in their host system or within a virtualenv they create. As a result, they end up needing to either install it in _this_ virtualenv (which means they need to activate this virtualenv first _anyway_) or learn the mastery of having it in another virtualenv without that virtualenv being fully active when they run things (yet having `pipenv` on `PATH`). The former of these has `pipenv` just be a complication over the regular python tools (i.e: `pip`) and the latter requires substantial Python & system admin-fu which it seems unreasonable to expect from casual contributors.

If we want to take the opinion that this repo wants to control the virtualenv which it runs in, then we should provide scripts and documentation around creating that, rather than assuming that the user knows how to do so. Until now, it hasn't done that. (If we did do this I would argue we should _still_ allow the user to have their environment somewhere else if they wanted, probably by having our scripts detect an existing env is active, so that people can use things like `virtualenvwrapper` if they want)

RealOrangeOne: This thread has gone to quite a tangent now, and isn't especially constructive to the PR.

I'd not noticed the documented setup instructions don't mention a virtualenv. I can live with the scripts staying there, and as the platform-agnostic commands are documented first, that should be fine.

I'd be interested in simply documenting that virtualenvs exist, and are considered best practice, so volunteers don't pollute their global environment.

PeterJCLaw:
https://github.com/srobo/runbook/pull/115 simplified things considerably. I do still think we're going to want something here, though this is no longer perssing.

PR #68, opened 2019-08-31
Experiment with converting the SVG flowcharts to mermaid

PeterJCLaw:
So far this just does the reception desk and I've intentionally left the original in place for comparison.

This (in my opinion) isn't actually as clear as the original, not least because it occupies much more space on the page. However it is now considerably easier to read the source for and should be a bit easier to update because you can focus on the logical changes.

Unfortunately the development experience around using mermaid here is pretty bad:
- you need to learn the `mermaid` syntax (not actually bad, just unique)
- your edit-refresh cycle is then also slow on two counts:
  - `mkdocs` reloading; which isn't particularly fast at this point and will likely only get slower as the runbook gets bigger
  - page reloading; especially bad if you're on medium or slow internet
- `mermaid` doesn't seem to have good error handling -- any cases where it doesn't like your input will partially render and then give up, without even putting a useful error into the debug console

I don't know if there's an online/realtime editor I could be using instead; that would help a lot, though moves the edit cycle even further towards needing to be online to do anything.

Anyway, here's some screenshots of the renderings:

Current SVG:
image

New via mermaid:
image

### DIFF
--- a/docs/competition/teams/role-descriptions/reception.md
+++ b/docs/competition/teams/role-descriptions/reception.md
@@ -38,7 +38,7 @@ graph TD
   q-have-mcf{"Do they have<br>a MCF?"}
   q-over-16{"Are they<br>over 16?"}
   fill-mcf[Get them to fill in a MCF]
-  wristband["Give the ma wristband<br>and attach it to them"]
+  wristband["Give them a wristband<br>and attach it to them"]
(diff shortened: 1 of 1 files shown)


Adimote:
Apologies for the drive-by comment: Right now the older flowchart still looks better IMO

PeterJCLaw:
Doesn't feel like mermaid is the best solution here.

PR #73, opened 2020-01-22
Add the Kit Team Vision

trickeydan:
This is the Kit Team's Vision. It was approved in a meeting on 2020-01-22.

This Pull Request also renames the "kit" section to "Kit Team", to reflect the separate section of SR. It would not make sense to have Kit Team as a subsection of Kit, as we are not robotics kit.


docs/kit-team/README.md
```diff
@@ -0,0 +1,100 @@
+# Kit Team
```
RealOrangeOne: RFC: I'm personally wary about renaming this section to "Kit Team", over simply "Kit". It feels strange to group everything under the team, rather than the domain, even though ~100% of what's considered "Kit" falls under the remit of the kit team.

In much the same way, renaming the "Competition" section "Competition Team" would also feel weird, if not more so, as almost everything else falls under that domain.

Personally, I'd have 2 tabs. 1 for the kit itself (as-is), 1 for the kit team organisations. But that's me.

(Intentionally not done as a review so as not to block merging, just registering concern. More than happy to be proved wrong!)

PeterJCLaw: Why not put this doc in the "Volunteering" section and leave the "Kit" section as it was?

rgilton:
There are two teams within SR, so it makes sense that the documentation is split along this organisational boundary -- without this split it will be exceedingly difficult to progress at any sensible speed.

docs/kit-team/README.md
TomWheal: My feeling towards this is the current runbook does not have the information about how either the compeition cycle ran last year, or how the kit has been managed the last couple of years. As such modifying the runbook to describe the new stucture of SR seems logical. 

The current content of the Kit Tab will likly all be written from scratch, as very little of it is now relevent to how the kit will be orginised intthe future. With the Assets team taking over the orginisation of the kit, and the develeopment being run by the relevent project team.

(Also on a side note, kit isnt really a good name for the current content either, as it describes the duties of a group of people whos purpose is to look after the kit, it doesnt talk much about the kit itself, which is the purpose of the docs)

WillB97: One thing to note is that having two levels of the hierarchy labelled as "team" makes it hard to understand how teams are grouped.
Labelling the project teams as sub-teams makes it clear they are grouped under the kit team rather than separate top level teams.

PeterJCLaw:
> There are two teams within SR, so it makes sense that the documentation is split along this organisational boundary

Yes, there are currently two main teams within the volunteers of SR. However I strongly disagree that this should be used as the basis for the documentation hierarchy.

There are a number of reasons for this.

(I'd hoped not to have this discussion on this PR, but oh well)

While we have two teams, the expectation and hope is that volunteers will be able to contribute broadly and wherever they feel able. Having everything separated by team will encourage isolation between people as well as soloing of knowledge, understanding and best practise. In short it will make it hard for the teams to work together, which they inevitably need to do for SR to be successful.

This leads to my second point: when there are topics of joint responsibility (e.g: helpdesk at the competition, communicating with teams regarding kit, among others) a hierarchy based on the teams leaves these topics without a clear home in the documentation. In turn this leads to discussions (such as this one) about where things go, rather than letting the focus be on making the docs the best they can be. An alternative approach of grouping things by when and how they happen avoids this completely -- for example helpdesk (somewhat obviously) fits under "competition". 

Finally we should remember that the Competition and Kit teams exist to make the competition happen and the kit exist respectively. As such, it is the competition and the kit which are themselves the primary entity, not the teams which exist to enable them. I would rather the runbook focus on the important _things_ rather than the (current) organisational constructs. This enables the runbook to be generally useful regardless of who is doing what, rather than needing rewriting all the time.

I'd definitely be interested in having a discussion for better structures for the runbook (the current is an amalgamation of a previous failed structure, rather than something deeply thought about), though I would like that to be a considered discussion based on the best form for the runbook and its general consumption within our open community than being pushed through as part of a PR which fundamentally doesn't need to make a structural change.

>  -- without this split it will be exceedingly difficult to progress at any sensible speed.

I don't understand why the speed at which the runbook can be updated has any relation to the renaming of the "kit" section as proposed here. Could you explain why this is a constraint?

docs/kit-team/README.md
PeterJCLaw: > My feeling towards this is the current runbook does not have the information about how either the compeition cycle ran last year, or how the kit has been managed the last couple of years. As such modifying the runbook to describe the new stucture of SR seems logical.

For SR2019 the handling of the kit was pretty close to what is currently documented, at least as far as kit collation and logistics go. I don't recall specifically referring to the packing docs, though I think we did something very similar to what is documented. For sure there are things we missed or did slightly differently (as is natural), but the intent was to follow the runbook's guidance. (It also didn't help that we were in the process of pulling the runbook together at the time)

> The current content of the Kit Tab will likly all be written from scratch, as very little of it is now relevent to how the kit will be orginised intthe future. With the Assets team taking over the orginisation of the kit, and the develeopment being run by the relevent project team.

This seems reasonable, though I'm not sure why this has any relation to the renaming of the section?

> (Also on a side note, kit isnt really a good name for the current content either, as it describes the duties of a group of people whos purpose is to look after the kit, it doesnt talk much about the kit itself, which is the purpose of the docs)

The section describes the general topic area within SR. I've outlined in [my comment](https://github.com/srobo/runbook/pull/73#issuecomment-578409259) a bit more about this, but the idea is that the runbook should document things not people†. It's somewhat unfortunate that it's currently framed around people, however this is due to a lack of time following the unification of the previous disparate docs.

† this gets a little complicated when consider the volunteering section, which is obviously also about people, but probably best thought about as "volunteering" being the _thing_ which is being documented. When viewed like that, the ways in which you can volunteer are therefore natural members of that topic.

trickeydan:
> While we have two teams, the expectation and hope is that volunteers will be able to contribute broadly and wherever they feel able.

I agree, as do many others within SR.

> Having everything separated by team will encourage isolation between people as well as soloing of knowledge, understanding and best practise.

By definition in the Ops Manual, there are aspects of SR that certain teams are wholly responsible for. However, there are equally things that are not well defined (e.g this runbook, infrastructure). 

What content within the new "Kit Team" section is not within the remit of the Kit Team? Leaving content scattered in a rather disorganised fashion as it currently is will only cause further confusion for volunteers.

> when there are topics of joint responsibility (e.g: helpdesk at the competition, communicating with teams regarding kit, among others) a hierarchy based on the teams leaves these topics without a clear home in the documentation.

Unless I made a lot of typos that I didn't notice, this PR does not introduce a hierarchy based on the teams. Content which is shared can still be in other sections. The content moved into the "Kit Team" section is either legacy docs which will be refactored in the future, or *content exclusive to Kit Team operations*.

> I would rather the runbook focus on the important things rather than the (current) organisational constructs

Where would you suggest we document these then? By this logic, I should close this Pull Request.

> than being pushed through as part of a PR which fundamentally doesn't need to make a structural change.

As far as I can tell, there are no other places for us to publish this document. Creating a separate PR for the structural would only delay the Kit Team launch further.

> I don't understand why the speed at which the runbook can be updated has any relation to the renaming of the "kit" section as proposed here. Could you explain why this is a constraint?

This Pull Request is a great example of this. Imagine if this took place in a company, whereby every single piece of content for internal docs is reviewed by every employee, each of who has the ability to block it. It is extremely difficult to reach consensus on everything in a group of more than about 5 people.[\[1\]][1] That's why voting was invented!

There needs to be a split, with maintainers for the sections of the runbook.

If we continue with a system where any volunteer can block things, we are likely to encounter the following problems:

- Preservation of the status quo
	- In decision-making bodies that use formal consensus, the ability of individuals or small minorities to block agreement gives an enormous advantage to anyone who supports the existing state of affairs.
	- We have seen in the recent years of SR, that keeping the status quo is not necessarily the right decision. Just because we've done it in the past, doesn't mean we still should.
- Susceptibility to widespread disagreement
	- Giving the right to block proposals to all group members may result in the group becoming hostage to an inflexible minority or individual.
- Stagnation and group dysfunction
	- When groups cannot make the decisions necessary to function (because they cannot resolve blocks), they may lose effectiveness in accomplishing their mission.
- Susceptibility to splitting and excluding members
	- When high levels of group member frustration result from blocked decisions or inordinately long meetings, members may leave the group, try to get to others to leave, or limit who has entry to the group.

[1]: https://web.archive.org/web/20060630154451/http://geocities.com/collectivebook/introductiontoconsensus.html

PeterJCLaw:
> > Having everything separated by team will encourage isolation between people as well as soloing of knowledge, understanding and best practise.
> 
> By definition in the Ops Manual, there are aspects of SR that certain teams are wholly responsible for. However, there are equally things that are not well defined (e.g this runbook, infrastructure).

This is true, though I'm not sure what it offers regarding the concern of isolation & siloing?

> What content within the new "Kit Team" section is not within the remit of the Kit Team? Leaving content scattered in a rather disorganised fashion as it currently is will only cause further confusion for volunteers.

Remits are not in question here, rather it's forcing the runbook to reflect the current remits are. We've tried that before and it didn't go particularly well (specifically while documentation was created and valuable it was hard to find). The runbook is (in part) an effort to fix that lack of discoverability by using a different organisational structure.

I completely agree that the current structure (and content) isn't perfect, though I'd rather work on fixing that holistically than start slicing it up piecemeal.

> > when there are topics of joint responsibility (e.g: helpdesk at the competition, communicating with teams regarding kit, among others) a hierarchy based on the teams leaves these topics without a clear home in the documentation.
> 
> Unless I made a lot of typos that I didn't notice, this PR does not introduce a hierarchy based on the teams. Content which is shared can still be in other sections. The content moved into the "Kit Team" section is either legacy docs which will be refactored in the future, or _content exclusive to Kit Team operations_.

Under the proposed change (that all Kit Team responsible things be under the Kit Team section) this PR has not moved anything incorrectly, rather the issue is that it _doesn't go far enough_. However I hope you'd agree that following through and moving all those other things into the Kit Team section would make things even less well structured (I seriously doubt any volunteer is going to think to look first for helpdesk under "Kit Team" -- it's clearly a competition thing). It's this result of things not being where you would first expect them to be that I see as one of the major issues.

> > I would rather the runbook focus on the important things rather than the (current) organisational constructs
> 
> Where would you suggest we document these then? By this logic, I should close this Pull Request.

Apologies, this perhaps isn't as clear as I'd hoped. I was referring here to the structure and approach used when adding content to the runbook. I'm definitely not suggesting that the new content in this PR doesn't belong here. As I noted originally, I think it would fit well in the "Volunteering" section.

> > I don't understand why the speed at which the runbook can be updated has any relation to the renaming of the "kit" section as proposed here. Could you explain why this is a constraint?
> 
> This Pull Request is a great example of this. Imagine if this took place in a company, whereby every single piece of content for internal docs is reviewed by every employee, each of who has the ability to block it.

This is not what is happening here. The reviews of the _content_ have been generally positive and not blocking. The blocking concerns here are around the changes to the _structure_ and how that affects the Runbook as a whole.

(With hindsight I wish I'd been able to write up more of my thoughts on the runbook's structure back when I originally pulled it together.)

I completely agree that having the Kit Team as the maintainers of the kit-focussed sections makes sense. For PRs within those sections, I expect that the Kit Team will be able to operate with considerable latitude and autonomy.

trickeydan:
Superseded by https://github.com/srobo/kitbook/pull/1

PR #76, opened 2020-01-25
Teams section update / cleanup

RealOrangeOne:
Go through and update the teams section for how we do things currently, along with consolidate and cleanup content.

The primary change was the removal of LTCs. I went with the option of removing things outright, although I'd be open to creating a page describing what they were, and that we don't operate things like that anymore.

PR #77, opened 2020-01-25
Add a page noting volunteer teams and their historical members

RealOrangeOne:
Add a list of members of the volunteer teams.

I'm not especially happy with the opening line. I'd like to add more emphasis that it's not just these volunteers who do everything in SR, and that every volunteer helps, but I couldn't find wording I was happy with. Input appreciated!

### DIFF
--- a/docs/volunteering/teams.md
+++ b/docs/volunteering/teams.md
@@ -1,0 +1,55 @@
+# Volunteer Teams
+
+Student Robotics has 2 teams which organise the operations regarding the wider competition programme. Each team is organised by its respective committee.
+
+- [Competition Team](https://opsmanual.studentrobotics.org/annual-robotics-competition/competition-team)
+- [Kit Team](https://opsmanual.studentrobotics.org/annual-robotics-competition/kit-team)
+
+## 2020
+
+### Competition Team
+
+##### Committee
+
+- Andy Barrett-Sprott
+- Andy Busse
+- Jake Howard
+
+##### Members
+
+- Antoine Petty
+- James Seden Smith
+- Jenny Fletcher
+- Peter Law
+
+### Kit Team
+
+##### Committee
+
+- Dan Trickey
+- Kajetan Champlewski
+- Kier Davis
+- Rob Gilton
+- Tom Wheal
+
+##### Members
+
+TBC
+
+## 2019
+
+In 2019, the Competition Team was known as the Core Team.
+
+### Core Team
+
+- Alistair Lynn
+- Andrew Barrett-Sprot
+- Andy Busse
+- Antoine Petty
+- Anton Nikitin
+- Dan Trickey
+- Holly Holder
+- Jake Howard
+- Kajetan Champlewski
+- Peter Law
+- Tyler Ward
--- a/docs/volunteering/teams.md
+++ b/docs/volunteering/teams.md
@@ -5,7 +5,7 @@ Student Robotics has 2 teams which organise the operations regarding the wider c
 - [Competition Team](https://opsmanual.studentrobotics.org/annual-robotics-competition/competition-team)
 - [Kit Team](https://opsmanual.studentrobotics.org/annual-robotics-competition/kit-team)
 
-## 2020
+## SR2020
 
 ### Competition Team
 
@@ -36,9 +36,9 @@ Student Robotics has 2 teams which organise the operations regarding the wider c
 
 TBC
 
-## 2019
+## SR2019
 
-In 2019, the Competition Team was known as the Core Team.
+During SR2019, the Competition Team was known as the Core Team.
 
 ### Core Team
 


trickeydan (dismissed):
I'm requesting changes on the grounds of data protection.

Please confirm that you have permission from individuals to list these names. 

What is the plan for removing the names should the individuals revoke their consent?

RealOrangeOne:
@TomWheal I was originally expecting this to contain a bit about the teams, but that duplicates stuff from the ops manual, which isn't ideal. Yeah moving it into the "Home" section might be better, can give it a look.

@trickeydan not something i'd considered. Revoking consent from git repos is something which no one has worked out, and generally falls under the "not technically feasable" category (most-merge, anyway). From a volunteer perspective, these names are already known, and a historical reference of them could be useful. An alternative, DP-friendly, implementation would be to drop the historical nature, and simply link to the GitHub teams, which enumerate the team and committee members, although I feel that loses something.

docs/volunteering/teams.md
```diff
@@ -0,0 +1,55 @@
+# Volunteer Teams
```
PeterJCLaw: Perhaps use a synonym for teams here if you're concerned about confusion versus competitor teams?
Maybe "organisation", "structure" or "groups"?

docs/volunteering/teams.md
```diff
@@ -0,0 +1,55 @@
...
+
+Student Robotics has 2 teams which organise the operations regarding the wider competition programme. Each team is organised by its respective committee.
+
+- [Competition Team](https://opsmanual.studentrobotics.org/annual-robotics-competition/competition-team)
+- [Kit Team](https://opsmanual.studentrobotics.org/annual-robotics-competition/kit-team)
+
+## 2020
```
PeterJCLaw: Is this the calendar year 2020 or the SR year 2020? It looks like it's meant to be the SR year (which the competition team uses), however we should probably use "SR2020" to clarify that. However of course the Kit Team uses different timings, so using SR years there might be confusing.

Are there other ways of presenting this information which avoids this conflict? (If you've considered some but this still turns out best. please do say!)

docs/volunteering/teams.md
RealOrangeOne: The issue is that said groups of people are named "Team", so "Teams" is the logical plural. 

See also https://github.com/srobo/runbook/pull/76#issuecomment-578442154

docs/volunteering/teams.md
RealOrangeOne: I'd intended this to be SR years, so yeah I think explicitly noting that is probably a good idea.

AFAIK, the Kit Team tenure falls conveniently around 2 competition cycles, so duplicating the list for the SR year is still correct. Thereby showing who was on the team for a given year, which is still useful information.

RealOrangeOne:
@TomWheal I tried moving the section over to under "Home", although it felt less useful there. The volunteer teams sit quite well under the "Volunteering" section, as we are all volunteers.

I wonder if a separate top-level section, something like "Organisation", which contains the high level organisation things about SR, such as the existence of the ops manual, this teams list, things like that? Although that's probably OOS for this PR. 

Personally it feels fine where it is (for now at least, anyway), although if more people tell me i'm wrong, I'm happy to move it :)

TomWheal:
Having had a bit of a think, I think ive identified why there might be such differing opinions on what the runbook should contain.
My understanding is:
The Ops Manual which describes how the charity operates IE top level teams, budgeting, sponsorship, how to get involved ETC.

The Runbook, details how those top level teams run, for example the way in which communicating with teams has been defined, or the structure of the kit team. 

Its quite possible with the creation of the ops. manual, the whole volunteering section should live on there, along with the content of this PR

RealOrangeOne:
> Its quite possible with the creation of the ops. manual, the whole volunteering section should live on there, along with the content of this PR

This PR shouldn't live on the ops manual, as the ops manual is a living document, rather than an archive. But yes it's definitely possible some bits of content should live there instead. The point of the volunteering section is to be a much less formal version of the ops manual, it obviously shouldn't contradict or duplicate. Ops manual is more about how the charity operates at a high level.

PeterJCLaw:
I'm conflicted on whether this belongs in the Runbook. On the one hand, we don't currently have a better place, yet on the other hand I'm not really sure what useful information this really provides. While in theory you could look at this to know who might know things, it doesn't really have enough detail for that.
I definitely agree this shouldn't be in the ops manual though.

There are also (small) GDPR concerns here about including names (I'm not enough of an expert to know for sure what the impact is, though I do think there's some conflict here -- people do need to be able to easily update their old info and opt-out of things), which might be enough to tip this from being slightly beneficial to not worth it.
I'm not sure we'd need permission from the individuals to list names - this is factual already-public information, being recorded for reasonable purposes. We would need to remove names if asked though. I am not a lawyer though.

trickeydan:
> I'm not sure we'd need permission from the individuals to list names - this is factual already-public information, being recorded for reasonable purposes.

Please could you link *where* this is already public. That is likely also a breach of GDPR.

Reasonable purposes would include holding the data internally, but would not cover publishing the names of past committee members.

trickeydan:
> Also @trickeydan I disagree, your name is already on GitHub being linked to the kit team

Yes, it is. However, GitHub does not currently list the names of previous committees / core team.

PeterJCLaw:
@RealOrangeOne this feels contentious enough (and GDPR skirting enough) that I suggest we drop this. I'm going to close this on those lines; if you feel strongly that you want this back then I suggest starting a discussion in Slack.

PR #78, opened 2020-01-26
Various syntax fixes & tidyup

PeterJCLaw:
This fixes various minor issues with syntax & style, mostly as highlighted by the VSCode extension I've added as a recommendation, though also some things which were causing rendering issues.

It would be great to get this reviewed & merged swiftly as I expect it'll generate conflicts _very_ quickly.

There shouldn't be any changes to content here; just formatting.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,5 +5,10 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+[*.md]
+# Needs to be 4 so that nested lists are properly interpreted by
+# Python-Markdown, which insists on four-space indents.
+indent_size = 4
+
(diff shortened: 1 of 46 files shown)


PeterJCLaw:
> I'd be interested in adding https://github.com/igorshubovych/markdownlint-cli to CI to ensure things remain standard, seems pointless to add a lint config if we don't actually lint it

:+1: though we're a long way from the linting passing. I'd missed that there was a CLI for it, that'll definitely make things easier!

PeterJCLaw:
I'm going to merge this now as I think this is a strict improvement _anyway_, but agree we should get a linter on CI at some point.

PR #80, opened 2020-02-08
Merge game design coordination into the game design section

PeterJCLaw:
This largely rewrites the section to reflect the process used for SR2019 and SR2020 as well as removing the obsolete content.

PR #81, opened 2020-02-08
Rework legacy role-based docs to be about the things they document

PeterJCLaw:
This updates much of the subsumed legacy docs to be closer to the usual approach of the runbook. I've focussed for now on the places where there were roles with people's names attached, mostly because those were all out of date.

This PR deliberately doesn't change places where the things being document still are roles -- for example the shepherding roles at the competition.

I've tried to keep the commits here roughly atomic, so reviewing them one at a time may be the easiest way to review this. I've also tried to avoid rewriting large chunks of documentation while making this change, preferring to change section introductions where the rest of the documentation still makes sense.

I've also tried to avoid moving things around here, though as a consequence of these changes the `competition/event` and `competition/production` sections are now a bit sparse and likely would be better served just being pages in the root of `competition` directly. I plan to do that separately after this merges. I've also handled the game design coordination move separately (#80).


docs/competition/README.md
```diff
@@ -0,0 +1,12 @@
...
+The culmination of Student Robotics annual [Competition Programme][programme] is
+the the Competition Event. This is typically a two-day event taking place on a
+weekend in April.
+
+During the course of the competition event the teams' robots compete in a large
+number of [matches](./matches/README.md) in order to determine the overall winner.
+
+This section outlines the running of the competition event.
+
+[programme]: ../programme/README.md
```
RealOrangeOne: Any reason to have this split out of the main text? Makes sense if it's used multiple times, but less so as it's just used once. Would be good to keep things consistent

docs/competition/event/README.md
```diff
@@ -1,25 +0,0 @@
...
-# Competition Event Documentation
-
-This is the Competition Event Coordinator's operations documentation.
-
-## Subordinate Roles
-
-Role | Volunteer | Documentation
------|-----------|--------------
-Reward Manager |
-Competition Volunteer Welfare Coordinator |  | [Docs](https://gist.github.com/howiegoing/5876963bc0dcb63a6fa1028b02d1a5a6)
```
RealOrangeOne: This item doesn't seem to be covered by another section. Worth importing / rewriting internally?

docs/competition/README.md
PeterJCLaw: Mostly for nice text wrapping -- this would have created odd looking line lengths on the line it's in.

docs/competition/event/README.md
PeterJCLaw: I believe this is superseded by https://studentrobotics.org/runbook/competition/volunteers/refreshments/. Is there content you feel is in the gist which is missing from that page?
(Note that our approach seems to have since changed from occasionally taking refreshments around to stationary posts to essentially always having roving refreshments)

RealOrangeOne: That page looks sufficient to me!

PR #84, opened 2020-02-23
Pin deps to versions specified in lockfile

RealOrangeOne:
Also remove dependency on `pymdown-extensions`, so it's handled by the underlying package which requires it (the theme).

Replaces #67, in a less intrusive way. Turns out this issue also prevented us installing things, as it did a full update of all dependencies, which sucked!

I forget exactly why I pinned everything to `*`, but now this shouldn't happen again.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -3,15 +3,15 @@ name = "pypi"
 url = "https://pypi.org/simple"
 verify_ssl = true
 
-[dev-packages]
-
 [packages]
 mkdocs = "==1.0.4"
 mkdocs-material = "==4.1.1"
 mdx-include = "==1.3.3"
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
@RealOrangeOne this seems to update a bunch of dependencies too -- is that expected?

PeterJCLaw:
@RealOrangeOne this seems to update a bunch of dependencies too -- is that expected?

RealOrangeOne:
> @RealOrangeOne this seems to update a bunch of dependencies too -- is that expected?

I don't _think_ so, but it doesn't appear to be affecting the final output. I suspect these are updates of loosely defined sub-dependencies. So it should still match the requirements they're after.

PR #85, opened 2020-02-23
Minify

RealOrangeOne:
PR based on #84, so it was actually possible to install things, I'd suggest not reviewing until that's merged.

Use https://github.com/byrnereese/mkdocs-minify-plugin to minify the site. This only minifies the HTML content, rather than any ancillary scripts.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,12 +6,11 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-mkdocs = "*"
-mkdocs-material = "*"
-pymdown-extensions = "*"
-mdx-include = "*"
-yamllint = "*"
-mkdocs-markdownextradata-plugin = "*"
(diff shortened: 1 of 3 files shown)


PeterJCLaw:
Interesting idea. Given that we're serving this via nginx with compression (gzip etc.) enabled, does this save us much?
There's maybe a small saving between GitHub Pages and the nginx box, though since the latter is caching anyway that's a fairly minor concern.

The trade-off here is that the generated HTML/JS source is somewhat less readable both in development and production.

RealOrangeOne:
There's still _some_ saving to the end user, but I agree it does make dev slightly more annoying, if you're viewing the raw HTML, that is. 

Will leave this open and think on it some more.

RealOrangeOne:
For the sake of easier debugging, and diminishing returns, I think we can leave this.

PR #86, opened 2020-02-28
Document the calendars we use

RealOrangeOne:
:ronseal:

### DIFF
--- a/docs/volunteering/calendars.md
+++ b/docs/volunteering/calendars.md
@@ -1,0 +1,15 @@
+# Calendars
+
+We use a few calendars to organise events, both public and private.
+
+## Volunteer Calendar
+
+We use a Google calendar to track events relevant to volunteers.
+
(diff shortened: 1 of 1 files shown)


docs/volunteering/calendars.md
```diff
@@ -0,0 +1,15 @@
+# Calendars
+
+We use a few calendars to organise events, both public and private.
+
+## Volunteer Calendar
+
+We use a Google calendar to track events relevant to volunteers.
+
+You can view and subscribe to this calendar [here](https://calendar.google.com/calendar/embed?src=studentrobotics.org_oqdjasvpps8smo0d5nte417rak%40group.calendar.google.com&ctz=Europe%2FLondon).
```
PeterJCLaw: As a general note: it's preferable to use some informative text as the text for a hyperlink, both for search engines and for accessibility reasons. Is there a way to word this which avoids using "here" for the link text?

(The later links in this document are great examples of how to do this better :+1:)

PR #88, opened 2020-03-14
Add section about task repo labels and what they mean

RealOrangeOne:
The tasks repo has some labels (https://github.com/srobo/tasks/labels) which have prefixes to categorize them. This was previously undocumented.

### DIFF
--- a/docs/volunteering/tasks.md
+++ b/docs/volunteering/tasks.md
@@ -48,6 +48,18 @@ please do create a task first as a record that you're doing it. There's no need
 to do this for trivial things, but for anything else having a task provides
 visibility into what's being done and who's doing it.
 
+## Labels
+
+The tasks repository has many [labels](https://github.com/srobo/tasks/labels) which are used to group tasks together, and assign them extra metadata.
+
+Labels in the repository are split into a few categories, denoted by a prefix character:
+
+- `A`: Area - The area of Student Robotics operations the task is relevant to.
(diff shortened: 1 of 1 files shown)


docs/volunteering/tasks.md
```diff
@@ -48,6 +48,18 @@ please do create a task first as a record that you're doing it. There's no need
...
 
+## Labels
+
+The tasks repository has many [labels](https://github.com/srobo/tasks/labels) which are used to group tasks together, and assign them extra metadata.
+
+Labels in the repository are split into a few categories, denoted by a prefix character:
+
+- `A`: Area - The area of Student Robotics operations the task is relevant to.
+- `I`: Importance - How critical the task is. Tasks may only have 1 `I` label at any given time
+- `S`: State - The current state of the task, which is generally timely.
```
PeterJCLaw: Do we allow tasks to have more than one `S` label at a time?

PeterJCLaw:
Nice, once we're happy with the wording I suggest duplicating into the readme in the tasks repo too, just because that's a bit closer to the tasks themselves. No need to get review on that if it's just a copy/paste of this change.

docs/volunteering/tasks.md
RealOrangeOne: Looking at the labels, I think we do.

For example, it's definitely valid for a task to be in progress and requiring signoff. There's a couple other examples of this.

PR #103, opened 2021-03-21
Update requirements

PeterJCLaw:
The primary motivation for this is to update Jinja2 for security, however since pipenv seems incapable of updating just one dependency at a time we end up doing all of them.

Since Python 3.5 is dead we also drop that here.

The `Pipfile.lock` change was generated by running `pipenv update` (using Python 3.8) and has been manually verified working in a Python 3.6 environment.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,6 +10,10 @@ mdx-include = "==1.3.3"
 mkdocs-markdownextradata-plugin = "==0.0.5"
 mkdocs-mermaid-plugin = {git = "https://github.com/pugong/mkdocs-mermaid-plugin"}
 
+# Something in 3.3 (specifically `(3.2.2, 3.3.4]`) breaks mermaid rendering.
+# Haven't looked into what just yet.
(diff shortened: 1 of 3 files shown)


PeterJCLaw:
Actually hold off on this, it looks like the mermaid stuff might not be working quite right.

RealOrangeOne:
I wonder if instead of doing this, we should take an approach similar to #67? It would bring the project in-line with our other packages, and make things significantly simpler to both develop and maintain.

PeterJCLaw:
Perhaps, though from what I recall that was rather controversial at the time. I think I have a fix for the mermaid issue anyway (pinning `markdown` for now).

PeterJCLaw:
For reference, http://localhost:8000/kit/logistics/asset-tracking/#asset-lifecycle is the page I'm using to test the mermaid rendering.

RealOrangeOne (approved):
Site builds, looks ok, and diagrams work fine.

PR #108, opened 2021-04-18
Set a start time for volunteers on day of livestream

sedders123:
Closes https://github.com/srobo/tasks/issues/656

### DIFF
--- a/docs/competition/virtual-competitions/competition-session-checklist.md
+++ b/docs/competition/virtual-competitions/competition-session-checklist.md
@@ -21,13 +21,13 @@ This checklist enumerates the tasks that need to be completed to run a virtual c
 - Revise the rules for the new module
 - Update the simulator for the new module in the private repo
 - Draft a post-event website article with details of the new module (any game sensitive content to be reviewed via Google Drive)
+- Set a volunteer start time for the day of the live stream (10am seems to work well for livestreams happening around lunchtime)
 
 ## After the Code Submission Deadline
 
 - Check for any no-shows
 - Construct the league for the session
 - Import the league into the compstate
-- Set a volunteer start time for the day of the live stream (10am seems to work well for livestreams happening around lunchtime)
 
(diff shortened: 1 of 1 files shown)


docs/competition/virtual-competitions/competition-session-checklist.md
```diff
@@ -27,6 +27,7 @@ This checklist enumerates the tasks that need to be completed to run a virtual c
 - Check for any no-shows
 - Construct the league for the session
 - Import the league into the compstate
+- Set a volunteer start time for the day of the live stream (10am seems to work well for livestreams happening around lunchtime)
```
PeterJCLaw, line 30: Is after the code submission deadline the right place for this? I know that's what we've tended to do, though I wonder if we should be doing it earlier?

PeterJCLaw (dismissed):

docs/competition/virtual-competitions/competition-session-checklist.md
RealOrangeOne: Definitely think we should be doing this earlier. I like the past ideas of trying to put points as early as possible in the list.

sedders123: I'm not 100% sure to be honest, my thinking on putting it here is that the start time is contingent on how well things go with league scheduling, match running etc. If anything goes wrong we may need to meet earlier if all goes correctly we could meet later?

PeterJCLaw: I think the goal of this entry was to provide the value for the less-involved volunteers, so that we have a date/time to publicise? I think it's fine if we determine that we actually need to change it.

sedders123: 26350e1

PeterJCLaw (approved):

RealOrangeOne (approved):

PR #115, opened 2021-05-17
Simplify dev setup, by removing `pipenv`

RealOrangeOne:
Replaces #67, but massively simplifies.

This is motivated by the issues around pipenv always upgrading every dependency whenever you touch any dependency.
(See https://github.com/pypa/pipenv/issues/2665 and https://github.com/pypa/pipenv/issues/2412)

Removing pipenv means fewer requirements for a development setup, along with a simpler and more-standard process.

I've also removed the YAML linter. There were no YAML files in the repo worth linting, and it's extra dependencies, test setup and confusion which isn't worth it or useful at this stage.

I've also elected not to use `pip-tools`. The dependency chain is very simple, and probably isn't worth the complexity at this point. If we find we need it, it'll be very simple to add back. We're already pinning everything we could reasonably need.


PeterJCLaw (approved):

trickeydan (commented):
It feels very dirty to do manual dependency resolution like this. I can foresee this change causing issues in the future, especially for less experienced volunteers.

RealOrangeOne:
> It feels very dirty to do manual dependency resolution like this. I can foresee this change causing issues in the future, especially for less experienced volunteers.

Manual dependency resolution? We're only pinning our direct dependencies (and the odd special case), the rest are done automatically at install time based on whatever the package itself says it requires. Yes it's less strict than tools with lockfiles, but personally i'd much rather that than the added complexity and annoyances which come with the other (less mature) tools out there.

This setup is much more widely used, and thus more likely to be familiar to volunteers with even a basic understanding of Python's packaging ecosystem. And if they have no experience with it at all, this is much simpler than both `pipenv` and `poetry`.

PeterJCLaw (approved):
(this hasn't changed since I last approved it)

PR #120, opened 2021-09-18
Add a proposal for review process guidelines

PeterJCLaw:
This is mostly a reflection of GitHub's terms, though with an explicit bias to action.

Frankly I'd hoped that much of this was obvious, though it seems not to be the case.

There are two things which this leads me to wonder about:
- Code Owners for restricting reviews on sensitive topics (safety/safeguarding)
- Automated merges (e.g: https://github.com/bors-ng/bors-ng) to enable _anyone_ to hit merge after approval

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,3 +9,40 @@ Contributions to Student Robotics' Runbook should:
 [code-of conduct]: https://opsmanual.studentrobotics.org/about-the-charity/code-of-conduct
 [license]: ./LICENSE
 [contribution-guide]: docs/contributing.md
+
+## Review Process
+
+We should aim to keep review as lightweight as possible, with a bias towards
+getting content live over making it perfect.
+
+We do however want to review the content which is contributed for two reasons:
+
+* content that is substantially unclear or actually wrong will do more harm than
+  a lack of documentation
+
+* some of the content here covers sensitive topics such as safety measures at
+  events or safeguarding; it is especially important that documentation of these
+  areas is correct & clear, warranting more scrutiny
+
+The following protocol (in line with how GitHub describe these terms) is
+therefore proposed for review:
+
+* _Approval_ responses mean "this change can merge as is" but may include
+  additional suggestions. Contributors may reject these suggestions, but are
+  expected to respond to them before merging if that is the case. Discussion of
+  these suggestions may continue after the PR is merged.
+
+* _Comment_ responses are non-blocking, but also not approval, typically to ask
+  questions about the content or suggest improvements. Such responses may
+  indicate a review only of some part of the content.
+
+* _Request changes_: responses are blocking and indicate that changes are needed
+  before content can be merged. Such responses should be rare.
+
+In general reviewers are encouraged to bias in favour of merging unless they
+feel that the change is particularly sensitive or would be a net negative to the
+runbook.
+
+After review the contributor is typically responsible for performing the merge
+themselves. New contributors (who may not have write access) should ping one of
+the maintainers or reviewers to do this instead.


CONTRIBUTING.md
```diff
@@ -9,3 +9,45 @@ Contributions to Student Robotics' Runbook should:
...
+  before content can be merged. Such responses should be rare.
+
+In general reviewers are encouraged to bias in favour of merging unless they
+feel that the change is particularly sensitive or would be a net negative to the
+runbook.
+
+### Practicalities of Review
+
+* Please do choose reviewers when opening a PR, ideally by subject matter but if
+  in doubt pick one of the maintainers and they can redirect
```
RealOrangeOne, line 49: If maintainers = code owners, their review will be requested automatically

If it's not, how will someone know who the maintainers are?

CONTRIBUTING.md
```diff
@@ -9,3 +9,45 @@ Contributions to Student Robotics' Runbook should:
...
+* Please do use GitHub's tooling to ask for re-review after changes
```
RealOrangeOne, line 50: Worth explicitly mentioning here that it's possible to request _teams_ for review, as opposed to individuals?

CONTRIBUTING.md
```diff
@@ -9,3 +9,45 @@ Contributions to Student Robotics' Runbook should:
...
+We do however want to review the content which is contributed for two reasons:
+
+* content that is substantially unclear or actually wrong will do more harm than
+  a lack of documentation
+
+* some of the content here covers sensitive topics such as safety measures at
+  events or safeguarding; it is especially important that documentation of these
+  areas is correct & clear, warranting more scrutiny
+
+The following protocol (in line with how GitHub describe these terms) is
```
RealOrangeOne, line 27: What happens if the majority of a review's comments are non-blocking, but a single point is. How is that reflected to the author?

Perhaps mentioning a standard like https://conventionalcomments.org/ would cover that?

CONTRIBUTING.md
PeterJCLaw: I was actually thinking of you & I as the maintainers of this and I'd originally thought that GitHub provided a way to see that about a repo, but it doesn't. Perhaps we should add that explicitly (somewhere)?

Yes Code Owners would be great to add, though I do expect that we'll have areas which may not have a Code Owner in the long term, hence the fallback to the maintainers.

CONTRIBUTING.md
PeterJCLaw: Perhaps, though it's not really clear to me what means -- is the implication that everyone in the team should review? I don't want this to get too far into "how to use GitHub" really, so I'd rather not to be honest.

CONTRIBUTING.md
PeterJCLaw: That's still blocking and it's be on the reviewer to clarify that. 9c6da4e encourages clarity here.

Conventional comments sounds like an interesting thing to try, however I'm concerned that these guidelines are already quite long! (I do separately wonder if we should have a set of generic guidelines for SR as a whole -- that could more easily be more detailed. Alternatively if we can build up a culture of using conventional comments then that alone may suffice)

CONTRIBUTING.md
PeterJCLaw: db4bbcc solves this for now.

CONTRIBUTING.md
PeterJCLaw: I'm going to leave this for now -- I think I'd rather solve this aspect via Code Owners or some similar mechanism rather than teaching people how to use GitHub.

trickeydan:
Is this intended exclusively for the runbook or for all SR projects?

PeterJCLaw:
> Is this intended exclusively for the runbook or for all SR projects?

As written this is intended for this repo only. There are definitely aspects which I think would be good to adopt everywhere (agreeing a protocol on what Approved/Request changes means and the expectations thereafter for example), but also things which likely don't make sense in other contexts (such as strongly biasing towards getting things in fast and fixing later, which would likely be a bad fit for firmware development).

PeterJCLaw:
> I have not read this yet but I suggest discussion/implementation of this is held until after the trustee organised meeting regarding documentation.

It's now been over a month since the trustees mooted the idea of such a meeting, yet nothing has happened. One of the core issues we have with documentation is a stagnation caused by allowing Perfect to be the enemy of the Good, which ironically is one of the things which these guidelines are aiming to resolve.

Unless you feel strongly that the proposal here is actively taking us in the wrong direction (bear in mind where we are currently), I suggest we consider this on its own merits and request review in that light. cc @RealOrangeOne @trickeydan

PR #122, opened 2021-10-03
Upgrade mkdocs and mkdocs-material

trickeydan:
Upgrades mkdocs and other dependencies, notably upgrading mkdocs-material to v7.

Also updated `mkdocs.yml` for slight changes in config. Some icons are slightly different due to the old icon-set no longer being available.

**Help Requested**: I am unable to work out why the menu structure doesn't render as intended.

Preview when rendered: https://static.trickey.io/runbook/

(I ended up on this adventure when attempting to do #2 as a quick thing)

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -4,12 +4,14 @@ repo_url: https://github.com/srobo/runbook
 strict: true
 edit_uri: blob/master/docs/
 theme:
-  logo:
-    icon: library_books
-  favicon: favicon.ico
-  name: material
-  feature:
-    tabs: true
+  name: 'material'
+  icon:
+    logo: fontawesome/solid/book
+  favicon: 'favicon.ico'
+  language: 'en'
+  features:
+    - navigation.tabs
+    - navigation.instant
   palette:
     primary: 'blue'
     accent: 'blue'
@@ -25,11 +27,11 @@ copyright: |
 # Customization
 extra:
   social:
-    - type: globe
+    - icon: fontawesome/solid/globe
       link: https://studentrobotics.org
-    - type: github-alt
+    - icon: fontawesome/brands/github-alt
       link: https://github.com/srobo
-    - type: twitter
+    - icon: fontawesome/brands/twitter
       link: https://twitter.com/studentrobotics
 
 # Extensions
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
-mkdocs==1.0.4
-mkdocs-material==4.1.1
-mdx-include==1.3.3
-mkdocs-markdownextradata-plugin==0.0.5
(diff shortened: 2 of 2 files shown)


mkdocs.yml
```diff
@@ -4,12 +4,14 @@ repo_url: https://github.com/srobo/runbook
...
-  feature:
-    tabs: true
+  name: 'material'
+  icon:
+    logo: fontawesome/solid/book
+  favicon: 'favicon.ico'
+  language: 'en'
+  features:
+    - navigation.tabs
+    - navigation.instant
```
PeterJCLaw, line 14: nit: Is this a required part of the upgrade?

RealOrangeOne:
> I am unable to work out why the menu structure doesn't render as intended.

I remember diving into this a while back. I don't believe it is possible anymore. It appears it broke around the time the navigation was rewritten in mkdocs-material (6.2.0). 

After digging around this again, I've submitted an issue upstream: https://github.com/squidfunk/mkdocs-material/issues/3076

mkdocs.yml
RealOrangeOne, line 14: Suggestion: We should add the [`navigation.index`](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#section-index-pages) feature. It means the dropdowns are clickable and have navigation, which will be incredibly helpful for some of the content we write.

PeterJCLaw:
Aside: I don't think your repro needs `navigation.instant`

mkdocs.yml
PeterJCLaw: Agree, though perhaps not as part of this upgrade PR?

PeterJCLaw (commented):
(I was digging into this too, but @RealOrangeOne beat me to an answer)

RealOrangeOne:
From the [related issue](https://github.com/squidfunk/mkdocs-material/issues/3076#issuecomment-932992826), it's no longer possible to have pages like we currently have them.

Personally I'd suggest we have a small meta section about the runbook itself which houses these page, or we attempt to bake them into he homepage page itself (that'll get quite long).

PeterJCLaw:
A breaking change removing something that is actually kinda hard to achieve another way occurring in a non-major release and then relegated to a footnote is rather disappointing IMO.

As I understand it the alternatives are:
- move the content to a meta folder and accept that it won't show up on the root page (plus maybe find some way to hide that folder from the tabs)
- some kind of redirect so that there is no actual root page and instead the root is `/meta` (would need to separately work out how to ensure that the meta directory's entry is the first tab in the list)
- fork the theme and/or override some aspect of it
- a plugin to adjust the nav (is that possible?)
- fully specify the _entire_ nav tree in the config

PR #124, opened 2021-10-03
Update packages

PeterJCLaw:
Extracted from #122 while that's blocked on nav issues.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-mkdocs==1.0.4
-mkdocs-material==4.1.1
+mkdocs==1.2.2
+mkdocs-material==4.6.3
 mdx-include==1.4.1
 mkdocs-markdownextradata-plugin==0.2.4
 mkdocs-mermaid2-plugin==0.5.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
 mkdocs==1.0.4
 mkdocs-material==4.1.1
-mdx-include==1.3.3
-mkdocs-markdownextradata-plugin==0.0.5
-mkdocs-mermaid2-plugin==0.5.1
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,8 +1,8 @@
 mkdocs==1.0.4
 mkdocs-material==4.1.1
```
RealOrangeOne, line 2: In testing, it should be safe to update both `mkdocs` to its latest, and `mkdocs-material` to a version lower than 6.2.0.

This will also require some configuration change, however.

PeterJCLaw: I've updated in https://github.com/srobo/runbook/pull/124/commits/85d2a0aec5f0d5099fd3f2e5b9474c174fbe9d50, though not quite that far. From what I can tell we don't need config changes with the updates I've got here.

PeterJCLaw:
I've diffed the built site and while there are some changes (notably dropping `title` attributes which repeat link contents) they seem unlikely to have meaningful impact.

trickeydan:
The "dict not hashable" errors are related to the config file not matching the expected schema. I got the same issues with v7 before fiddling with it.

PeterJCLaw:
Ah, interesting. I did look in the release notes and didn't spot anything, though perhaps I wasn't looking well enough.

PeterJCLaw:
Eurgh, [mkdcos-material's release notes for v5](https://squidfunk.github.io/mkdocs-material/changelog/#500-_-april-7-2020) completely don't mention the breaking change :(

PeterJCLaw:
In any case I think it's probably easiest to leave #122 to pick up the configuration migration, since that PR already has that mostly done?

sedders123 (approved):

PR #125, opened 2021-10-06
Update Kit Software section

trickeydan:
- Remove information about Legacy Server Maintainer role
  - This role no longer exists and keeping the documentation is confusing.
  - We can refer to the git history if we need it again.
  - It is not Kit Software
  - Also removed original author metadata as all previous content was removed.
- Add information about SR2022 kit stack


docs/kit/software/README.md
```diff
@@ -1,16 +1,79 @@
...
+
+It is maintained by the j5 Steering Committee.
+
+## Off-The-Shelf Components
+
+The following components are not maintained by SR and we do not substantially contribute to them.
+
+### Mosquitto MQTT Broker
+
+We use Mosquitto as a MQTT broker on the robots. This is a requirement for Astoria.
```
PeterJCLaw, line 69: Maybe worth a link to Mosquitto MQTT (like how you have for Arch; if there is a useful homepage link for the project)?

PeterJCLaw (dismissed):

docs/kit/software/README.md
trickeydan: Ah good spot, must have missed that one!

trickeydan: 96c1338

PR #129, opened 2021-11-09
Add definition of KLI to glossary

trickeydan:
We've found ourselves using the acronym KLI frequently for this, probably best to document it somewhere.


PeterJCLaw (approved):
:shipit:

PR #132, opened 2021-11-20
Add the start of an offboarding checklist

PeterJCLaw:
This is likely very incomplete, however provides a place for us to start adding things as we discover them.

The scope here is really the checklist of services, so it is explicitly _not_ the scope of this PR to make this checklist a complete guide for offboarding. (That would be great to do, but separately).

Fixes https://github.com/srobo/infrastructure-team-minutes/issues/12

### DIFF
--- a/docs/volunteering/offboarding-checklist.md
+++ b/docs/volunteering/offboarding-checklist.md
@@ -14,6 +14,9 @@ their teams' committees to each run through this checklist.
   important they know that they're welcome back and we're grateful for the
   support they have given us in the past.
 
+- Arrange for the volunteer to return any equipment which belongs to the
+  charity.
+
 - If possible, ask why they're leaving. This should be conversational to build
   understanding, it's not about trying to convince them to stay.
 
--- a/docs/volunteering/offboarding-checklist.md
+++ b/docs/volunteering/offboarding-checklist.md
@@ -1,0 +1,19 @@
+# Offboarding Checklist
+
+From time to time volunteers may decide that they would like to leave the
+organisation to focus on other things. When this happens we should try to ensure
+that anything they're responsible for is handed over to another volunteer.
+
+This checklist aims to provide prompts for things which might need to be handed
+over.
+
+## Third Party Services
+
+- Check [the list][list-third-party-services] of who is configured as the
+  "owner" of each third party service that we use and transfer as needed.
+
+- Check whether there are any other services which the volunteer manages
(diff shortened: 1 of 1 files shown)


trickeydan:
Drive-by thought: We should probably be asking for equipment back too?

Tyler-Ward:
We should also consider different levels of offboarding e.g. fully leaving vs taking a less active role or a single small role e.g. mentoring only

PeterJCLaw:
> We should also consider different levels of offboarding e.g. fully leaving vs taking a less active role or a single small role e.g. mentoring only

Definitely, though for now I'm trying to keep this simple so as to get the basics in place.

-----

@sedders123 @antoinepetty adding you as reviewers here as you've mostly been doing the onboarding of volunteers recently so may have some thoughts.

ping @dianedowling if you have thoughts too.

I'd quite like to get this in soon, so that we have a record of the linked spreadsheet and so that we have something in place to start improving.

sedders123:
Is the intent of this PR to replace the [Volunteer Management](https://github.com/srobo/ops-manual/wiki/Volunteer-Management) part of the ops-manual wiki? Seems like we may be making two places to look for information on offboarding. Personally, I think the runbook is likely the correct place for this but there is a lot of info in the wiki already (though it could well be outdated as our processes have changed). Do we want to copy the info from the wiki here or are we intentionally starting again?

PeterJCLaw:
Huh, I wasn't aware/had forgotten about that. My goal here was primarily to have a place to put the link to the services spreadsheet and it kinda snowballed a bit from there. There's definitely a lot of overlap here.

As part of the Infrastructure Team's responsibility to ensure access to services we probably need to have input into the off-boarding procedures, though that may only need to be something like "check the sheet to see what services the person needs to hand over".

I'll review those docs more and maybe put together an ops-manual PR.
Agree that maybe some of that might fit better in the runbook.

PeterJCLaw:
For visibility: I've raised the interaction with the ops-manual in https://github.com/srobo/ops-manual/issues/21, which the trustees are going to review (along with the other open ops manual issues) in their next meeting.

PeterJCLaw:
The conclusion on the ops manual seems to be that:
- this sort of content is a good fit for the runbook, so let's keep it here
- the wiki isn't something the current trustees were aware of; they're going to review it but likely it'll change somewhat

I therefore suggest that we move forwads with this PR as starting the new home for this sort of information and if things change we can update it to reflect that.

sedders123 (approved):

PR #137, opened 2022-02-01
Add docs on requesting infrastructure

PeterJCLaw:
This includes guidance on who to request infrastructure from as well as how to do so.

Fixes https://github.com/srobo/infrastructure-team-minutes/issues/14

### DIFF
--- a/docs/infrastructure/README.md
+++ b/docs/infrastructure/README.md
@@ -62,11 +62,6 @@ repositories is an [ongoing project][legacy-repository-migration].
 For our email (`@studentrobotics.org`) accounts as well as document sharing we
 use Google Workspace. Google Groups are used for various mailing lists.
 
-**Note:** The Infrastructure Team do not yet have administrative access to
-Google Suite, so account management requests must be made to the trustees
-instead. The Infrastructure team will be taking this over at some point in the
-future.
-
 ## Mailing lists
 
 For sending some batch emails we use Mailchimp and Google Groups.
--- a/docs/infrastructure/README.md
+++ b/docs/infrastructure/README.md
@@ -62,9 +62,10 @@ repositories is an [ongoing project][legacy-repository-migration].
 For our email (`@studentrobotics.org`) accounts as well as document sharing we
 use Google Workspace. Google Groups are used for various mailing lists.
 
-**Note:** As of early 2022, the Infrastructure Team do not have administrative
-access to Google Suite, so account management requests must be made to the
-trustees instead.
+**Note:** The Infrastructure Team do not yet have administrative access to
+Google Suite, so account management requests must be made to the trustees
+instead. The Infrastructure team will be taking this over at some point in the
+future.
 
 ## Mailing lists
 
--- a/docs/infrastructure/README.md
+++ b/docs/infrastructure/README.md
@@ -62,6 +62,10 @@ repositories is an [ongoing project][legacy-repository-migration].
 For our email (`@studentrobotics.org`) accounts as well as document sharing we
 use Google Workspace. Google Groups are used for various mailing lists.
 
+**Note:** As of early 2022, the Infrastructure Team do not have administrative
+access to Google Suite, so account management requests must be made to the
+trustees instead.
+
 ## Mailing lists
 
 For sending some batch emails we use Mailchimp and Google Groups.
--- a/docs/infrastructure/requesting-infrastructure.md
+++ b/docs/infrastructure/requesting-infrastructure.md
@@ -21,6 +21,22 @@ Team's budget to which that spending would relate.
 
 [slack-infra]: https://studentrobotics.slack.com/messages/infrastructure
 
+## Container hosting
+
+In the future the Infrastructure Team hope to offer a shared hosting platform
+for small services. These services will need to be provided as Docker
+containers, however the exact details (both technical and budgetary) of this
+setup have not yet been worked out.
+
+For now we encourage the use of containers for ease of deployment and the
+Infrastructure Team can help in running the host machine, but note that you will
+need to budget for your own machine for the containers to run on.
+
+Think of this as the "contact our Sales team for pricing" option where the other
+items listed here are much more "off the shelf" offerings, so please allow for
+some discussion around how container hosting will work when planning your
+services.
+
 ## GitHub
 
 When requesting changes related to GitHub, please include the GitHub username of
--- a/docs/infrastructure/requesting-infrastructure.md
+++ b/docs/infrastructure/requesting-infrastructure.md
@@ -43,6 +43,10 @@ When requesting changes related to GitHub, please include the GitHub username of
 the users affected (if applicable) as well as the names of any teams or repos
 those people will need access to.
 
+Note that in general you are encouraged to add individuals to teams and grant
+teams access to repos (rather than adding individuals to repos) as this makes it
+easier to maintain consistency.
+
 ## Google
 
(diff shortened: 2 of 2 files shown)


docs/infrastructure/requesting-infrastructure.md
```diff
@@ -0,0 +1,59 @@
...
+Infrastructure Team by starting a new conversation thread in the
+[`#infrastructure` channel][slack-infra] in Slack. Please ping
+`@infrastructure-committee` to ensure that your message is seen.
+
+In general you can expect a response within a couple of days (often much
+sooner), though note that the team may not always be able to act on your request
+for some time.
+
+Requests for new infrastructure must be made by (or confirmed by) members of a
+Team Committee.
```
PeterJCLaw, line 16: Should this be the budget holder on each committee? Is that worth it? (Do other people know who that is?)

RealOrangeOne, line 16: Nit
```suggestion
Team's Committee.
```

docs/infrastructure/requesting-infrastructure.md
```diff
@@ -0,0 +1,59 @@
...
+Where infrastructure requests would incur additional spending (for example a new
+machine from Mythic Beasts), you must include a reference to the line in the
+Team's budget to which that spending would relate.
+
+[slack-infra]: https://studentrobotics.slack.com/messages/infrastructure
+
+## GitHub
+
+When requesting changes related to GitHub, please include the GitHub username of
+the users affected as well as the names of any teams or repos those people will
```
RealOrangeOne, line 27: > the users affected

What if it's a (github) team?

docs/infrastructure/requesting-infrastructure.md
```diff
@@ -0,0 +1,59 @@
...
+For new resources please include in your request:
+
+- Kind of resource (VM, bare metal, DNS entry, etc.)
+- Level of urgency
+- Spec level (please use Mythic's order forms to determine what you want and use
+  their terminology when making requests)
+- Expected cost (as a check to help ensure the right options are selected)
+- Purpose/name
+- Operating System
+- SSH keys for bootstrap access (or where to get them)
```
RealOrangeOne, line 49: Personally I think assuming using the keys on their GitHub account is probably fine, to save duplication. Specifying who should be given access as part of the bootstrap though would be good.

docs/infrastructure/requesting-infrastructure.md
PeterJCLaw: A "Team Committee" is a thing too, so I think this is fine as is.

docs/infrastructure/requesting-infrastructure.md
PeterJCLaw: I don't disagree, however I'd much rather have a system where the person says "please use these keys" than one where we assume that we know their github username and that those are the keys they want us to use. There's also the case where a committee member is making the request but another team member is going to be administering the box.

docs/infrastructure/README.md
```diff
@@ -62,6 +62,10 @@ repositories is an [ongoing project][legacy-repository-migration].
 For our email (`@studentrobotics.org`) accounts as well as document sharing we
 use Google Workspace. Google Groups are used for various mailing lists.
 
+**Note:** As of early 2022, the Infrastructure Team do not have administrative
```
Tyler-Ward, line 65: This implies it will be the state going forward rather than a temporary limitation.
A potential better wording would be.

**Note:** The Infrastructure Team do not yet have administrative access to Google Suite, so account management requests must be made to the trustees instead. The Infrastructure team will be taking this over at some point in the future.

docs/infrastructure/requesting-infrastructure.md
PeterJCLaw: How about:
> When requesting changes related to GitHub users, please include the GitHub
> username of the users affected as well as the names of any teams or repos those
> people will need access to.

I'm not sure how to frame the teams thing without it just sounding like wasted words. Open to suggestions.

docs/infrastructure/requesting-infrastructure.md
```diff
@@ -0,0 +1,59 @@
...
+
+When requesting changes related to GitHub, please include the GitHub username of
+the users affected as well as the names of any teams or repos those people will
+need access to.
+
+## Google
+
+**Note:** As of early 2022, the Infrastructure Team do not have administrative
+access to Google Suite, so account management requests must be made to the
+trustees instead.
```
Tyler-Ward, line 34: same comment as in the readme

docs/infrastructure/requesting-infrastructure.md
Tyler-Ward: I believe other funding requests can be sent to the trustees by any committee member so any of them requesting infrastructure also seems reasonable

docs/infrastructure/requesting-infrastructure.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Mythic Beasts
```
Tyler-Ward, line 36: We mentioned about creating the ability to request container hosting given the other notes in here do we want to add a note that we plan to offer this.

Tyler-Ward (commented):
A few suggestions added inline

docs/infrastructure/requesting-infrastructure.md
Tyler-Ward: Nevermind the above just seen the email which was different than the meeting discussion

docs/infrastructure/README.md
PeterJCLaw: Yup 👍. I was struggling with a better wording, your suggestion is good.

PeterJCLaw: 54ec131

docs/infrastructure/requesting-infrastructure.md
PeterJCLaw: :+1: thoughts on 5f4f77e until we get this sorted?

docs/infrastructure/requesting-infrastructure.md
PeterJCLaw: f4b8979

docs/infrastructure/requesting-infrastructure.md
PeterJCLaw: Yeah, that was my reasoning here. I think from our perspective sign-off from a committee member should suffice, especially to avoid that becoming too much of a bottleneck. I'm going to suggest we run with this approach for now and re-evaluate if it either becomes an issue or the trustees feel strongly otherwise (cc @dianedowling for visibility).

PeterJCLaw:
For visibility: I've added https://github.com/srobo/runbook/pull/137/commits/84663c3d06d8e02f19d9585974c51987c6109049 following offering the same advice in Slack.

docs/infrastructure/requesting-infrastructure.md
Tyler-Ward: That looks good to me

Tyler-Ward (dismissed):

sedders123 (dismissed):

docs/infrastructure/requesting-infrastructure.md
PeterJCLaw: Confirmed in this evening's all committees meeting -- @dianedowling is happy for team-level budget signoff to be by committee members (as long as the committee in question is ok with that).

RealOrangeOne (approved):

PR #140, opened 2022-03-13
Add spellchecking in CI

PeterJCLaw:
Also encourage use of a compatible spellchecker in VSCode too.


.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,24 @@
...
+    branches:
+      - "*"
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
```
sedders123, line 14: v3 exists and as this is being introduced for the first time would be nice to start modern

```suggestion
        uses: actions/checkout@v3
```

sedders123 (approved):

.github/workflows/spellcheck.yml
PeterJCLaw: Ha, that must have happened since I created this!

PR #141, opened 2022-03-19
Collate and sort kit documentation

trickeydan:
Discussed in doing call today


Tyler-Ward (approved):

sedders123:
Was the Battery Charging section of `docs/competition/team-support/README.md` intentionally removed? Battery Shuffler has been left in and I'm not sure that makes sense without us doing the battery charging

PR #148, opened 2022-04-10
Update instructions on handling fire-y batteries

Scarzy:


Scarzy:
Ergh, this has brought in more than I expected.

PR #149, opened 2022-04-10
Update instructions on how to deal with a fire-y LiPo

Scarzy:
Based on feedback from Surrey Fire & Rescue Service

### DIFF
--- a/docs/kit/event-operations/battery-charging.md
+++ b/docs/kit/event-operations/battery-charging.md
@@ -46,6 +46,7 @@ The biggest risk of damage or injury is from the banana plugs coming unplugged f
 The batteries themselves have many built-in safety features and, unless abused, are completely safe. All batteries undergoing charging must be placed inside a fireproof charging bag. In the extremely unlikely event of a fire due to a battery:
 - DO NOT use a fire extinguisher - the batteries contain hazardous materials which cannot be safely tackled by most fire extinguishers.
 - DO NOT open or attempt to remove anything from the charging bag.
+
 ONLY IF SAFE TO DO SO:
 - A fire bucket full of water will be supplied, get the battery into the bucket. This should be performed complete with the charging bag, and any other contents of the charging bag. This battery must not be removed from the bucket.
 - Remove power from all other batteries and chargers, and move all undamaged equipment away from the faulty battery/charger.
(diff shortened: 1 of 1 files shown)


trickeydan (approved):

docs/kit/event-operations/battery-charging.md
```diff
@@ -43,7 +43,14 @@ Every volunteer performing the Battery Charging role must have read this documen
...
 
-The batteries themselves have many built-in safety features and, unless abused, are completely safe. All batteries undergoing charging must be placed inside a fireproof charging bag. In the extremely unlikely event of a fire due to a battery DO NOT attempt to tackle the fire - the batteries contain a flammable metal (Lithium) which cannot be safely tackled by most fire extinguishers.
+The batteries themselves have many built-in safety features and, unless abused, are completely safe. All batteries undergoing charging must be placed inside a fireproof charging bag. In the extremely unlikely event of a fire due to a battery:
+- DO NOT use a fire extinguisher - the batteries contain hazardous materials which cannot be safely tackled by most fire extinguishers.
+- DO NOT open or attempt to remove anything from the charging bag.
+ONLY IF SAFE TO DO SO:
```
WillB97, line 49:
```suggestion

ONLY IF SAFE TO DO SO:
```
I assume the intention is for these to be separate lists

TomWheal:
You can buy a fireproof box designed for bad lipos. Might be worth us getting one to put lipos in the bag inside in case of an emergency. 

https://uk.rs-online.com/web/p/battery-storage-boxes/2000580
(I know it's pricey)

docs/kit/event-operations/battery-charging.md
Scarzy: It is a continuation of the list "In the extremely unlikely event of a fire due to a battery:", but with the extra qualifier of only if safe, so I'm not sure if the extra line is helpful or confusing?

Scarzy:
> You can buy a fireproof box designed for bad lipos. Might be worth us getting one to put lipos in the bag inside in case of an emergency.

The intent of these instructions is for use when tackling one actually on fire, and is based on feedback from a fire and rescue service.
These boxes seem designed for transporting lipos that aren't fire-y, but not far off, and containing a fire should one break out. I suspect there are specific instructions on how to pack a battery in there, and it isn't just "throw and slam".
I'm really not comfortable with the idea of suggesting we know better than a fire and rescue service and in doing so using something in a manner it doesn't seem designed for...

Maybe there is an argument for getting one of those boxes, but I'm not sure it's applicable to this PR.

docs/kit/event-operations/battery-charging.md
WillB97: Without the new line the heading is appended to the last line of the previous list.

PR #152, opened 2022-04-12
Add a definition of a Doing

PeterJCLaw:


RealOrangeOne (approved):

WillB97:
Much review time, many wow.

This definition is essentially meaningless, it's like saying "walking is when you walk".
Something like "An event where volunteers work together to progress the tasks of the charity. i.e. preparing for the competition or testing the kit."

PeterJCLaw:
> Much review time, many wow.

Indeed, contributions to the runbook should be able to be made swiftly.

> This definition is essentially meaningless, it's like saying "walking is when you walk".

I disagree -- the point here is that Doings are pretty much anything that's not discussion.

> Something like "An event where volunteers work together to progress the tasks of the charity. i.e. preparing for the competition or testing the kit."

Yup, agree that this is a clearer definition.

PR #155, opened 2022-04-16
Add vague instructions on discharging batteries

Scarzy:
This was done in the absence of a charger, hence vagueness. Is currently more of a reminder that it needs to be done.


WillB97 (approved):
We may want to mention the current setting but since we don't change this anymore it's basically covered by the charging section.

PR #159, opened 2022-04-17
Update Kit Return protocol for SR2022

Adimote:


docs/kit/event-operations/kit-return.md
```diff
@@ -4,16 +4,18 @@
...
 
-### Missing Items
+When fully checked, the supervisor should fill out the [Kit Return Form](https://docs.google.com/document/d/1eegZOBdhhDcwjxq9rwLbtxBbY3gxckNpkNXDLWMA8Yo/edit?usp=sharing) to confirm everything is complete.
 
-If a kit is not complete, the [Missing Kit Form](https://github.com/srobo/kit-coordination-documents/tree/master/missing-kit-form) should be completed noting which items of kit have not been returned. After the competition, these forms are tabulated and the teams contacted to ensure the safe return of the items of kit.
+### Batteries
 
-[Example email for teams with outstanding single items of kit](https://github.com/srobo/team-emails/blob/master/SR2019/2019-05-28-single-item-kit-chase.md)
+As we handle charging during the competition, teams will have 1 Battery. We don't want to keep those in the kits as we want to discharge them and place them with all other batteries. Please remove the battery from the kits, check it off the checklist, and place them in a container to be taken back down to battery charging.
```
raccube, line 13:
```suggestion
As we handle charging during the competition, teams will have 1 Battery. We don't want to keep those in the kits as we want to discharge them and place them with all other batteries. Please remove the battery from the kits, check it off the checklist, and place them in a container to be taken back down to the battery charging area to be storage charged.
```

raccube (changes requested):
Looks good, just one thing to make things more clear.

raccube (approved):

PR #164, opened 2022-06-02
Update some definitions in the glossary

antoinepetty:
Removed old definitions and added new ones

### DIFF
--- a/docs/volunteering/glossary.md
+++ b/docs/volunteering/glossary.md
@@ -24,11 +24,6 @@ teams build their robots; see also _SRnnnn_
 : the group of volunteers responsible for delivering a particular year's
 _Competition programme_
 
-_Core Team_:
-
-: the group of 8 or so volunteers responsible for delivering the _SR2019_
-_Competition programme_; this has since been replaced by the _Competition Team_
-
 _Doing_:
 
 : an event where volunteers work together to progress the tasks of the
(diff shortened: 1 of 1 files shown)


samjmartin04 (approved):

docs/volunteering/glossary.md
```diff
@@ -24,11 +24,6 @@ teams build their robots; see also _SRnnnn_
 : the group of volunteers responsible for delivering a particular year's
 _Competition programme_
 
-_Core Team_:
-
-: the group of 8 or so volunteers responsible for delivering the _SR2019_
-_Competition programme_; this has since been replaced by the _Competition Team_
-
```
PeterJCLaw, line 31: If we're removing this definition please could we also remove all mentions of the term?

antoinepetty: Looks like there was only one so have removed it :)

docs/volunteering/glossary.md
```diff
@@ -56,10 +51,13 @@ _Mentor_:
...
-_Shepherd_:
+[_Pull Request (PR)_](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests):
+
+: the review stage of some changes on GitHub to code or documentation
+
+_Retro_:
 
-: a volunteer at the competition event who helps ensure teams make it to their
-matches on time
+: a meeting after an event to discuss the positives, negatives, feedback, and ways to improve for the following year
```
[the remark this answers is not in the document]
antoinepetty: It's quite a software industry specific term so I think we should document it if we're using it frequently to make sure everything's as accessible as possible

sedders123 (approved):

raccube (approved):

PR #166, opened 2022-08-01
Move to GitHub Actions for build & deploy CI

PeterJCLaw:
Now that there are actions for publishing to Pages this should work better for us.

This also inlines the spellchecking so we can depend on that before deploy.

Based on https://github.com/actions/starter-workflows/blob/main/pages/hugo.yml

The deploy failure here was due to an earlier (since force-pushed away) version which didn't have the right check to not deploy from a branch. I'm hoping that that will go away when this is run on `master` and we configure the repo accordingly (see https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site).


sedders123 (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,86 @@
...
+      - spellcheck
+      - build
+
+    if: github.ref_name == 'master'
+
+    steps:
+
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v1
```
RealOrangeOne, line 86: Nit: Would be nice to collapse some whitespace

```suggestion
    if: github.ref_name == 'master'
    steps:
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v1
```

RealOrangeOne (approved):

.github/workflows/main.yml
PeterJCLaw: https://github.com/srobo/runbook/pull/166/commits/555e3d8b7ace4809e7abab2fcdf039300bbfcbac
The space around the `if` is somewhat deliberate.

PR #170, opened 2022-10-26
kit/software: Update for SR2023

trickeydan:
Also have split into layers rather than split by who wrote the software.


raccube (changes requested):
Looks good, just one minor thing.

raccube (approved):

PR #171, opened 2022-10-26
Remove Out of Date Info and Update Kit Hardware Section

trickeydan:
- Removed information that is out of date and has no value other than in git history
- Renamed Kit Debugging to Hardware Debugging


PeterJCLaw (approved):

docs/kit/hardware/kit-definition.md
```diff
@@ -1,57 +0,0 @@
...
-  authors: Richard Barlow
-  url: https://bitbucket.org/richardbarlow/sr-kit-coord/wiki/Kit%20Definition
----
-# Kit Definition
-
-This page serves as the canonical list of parts constituting a Student Robotics kit.
-
-## Hardware
-
-Part                     | Inventory Asset Type  | Qty |
```
WillB97, line 12: We should probably keep this. This is the most complete list of what is in a kit (inc. things like the gauge and length).

PR #172, opened 2022-10-26
Move Kit Event Operations into Competition section

trickeydan:


PeterJCLaw (approved):

PeterJCLaw:
Just a thought (sorry I missed this earlier) -- should we have redirects from the old pages to the new? I think there's a way to specify this in the new pages, so the editing experience is still fairly simple.

PR #175, opened 2023-01-21
First pass at a general events planning checklist

PeterJCLaw:


prophile (approved):

WillB97 (approved):
We should probably add a comment somewhere noting that there will be more to do beyond the announcement.

PeterJCLaw:
> We should probably add a comment somewhere noting that there will be more to do beyond the announcement.

585f071dc7f26297403d310c893dbb150bce931a aims to help with that.

PR #182, opened 2023-03-17
Alternate roles for stopping the competition

WillB97:
The head judge should be a role of adjudication not organisation. Head shepherd and the lead of event tech (lighting, construction, etc.) are better placed to assess an incident.


samjmartin04 (approved):

RealOrangeOne (approved):

sedders123 (approved):
Personally I think anyone should be able to stop the competition, but if we're limiting I agree with these changes

prophile:
Wait what? This has been needed exactly once in practice, _by the head judge_.

PR #195, opened 2023-08-05
Add an initial page on Information Security

PeterJCLaw:
Mostly so that we can record a policy on using multi-factor authentication.

Also mentions server security, copying & linking content from where this was already covered.

PR for visibility before adopting this policy more formally.

### DIFF
--- a/docs/infrastructure/security.md
+++ b/docs/infrastructure/security.md
@@ -27,3 +27,20 @@ Useful links:
 - [Slack two factor authentication](https://slack.com/intl/en-gb/help/articles/204509068-Set-up-two-factor-authentication-Set-up-two-factor-authentication)
 
 [wikipedia-mfa]: https://en.wikipedia.org/wiki/Multi-factor_authentication
+
+## Servers
+
+<!-- Updating this? Also update "Securing the OS" in ./machine-setup.md -->
+
+Exactly what is needed to secure a given server will depend on its intended
+use-case, operating system and other factors. In general though we expect that:
+
+* the firewall will block everything that's not needed
+* root SSH is disabled
+* password SSH is disabled (i.e: keys only)
+* individuals have their own user accounts
+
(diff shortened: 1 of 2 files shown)


docs/infrastructure/security.md
```diff
@@ -0,0 +1,46 @@
...
+<!-- Updating this? Also update "Securing the OS" in ./machine-setup.md -->
+
+Exactly what is needed to secure a given server will depend on its intended
+use-case, operating system and other factors. In general though we expect that:
+
+* the firewall will block everything that's not needed
+* root SSH is disabled
+* password SSH is disabled (i.e: keys only)
+* individuals have their own user accounts
+
```
RealOrangeOne, line 42: Suggestion: explicitly state root access should be restricted

```suggestion
* individuals have their own user accounts
* root access is limited to only users who absolutely need it

```

RealOrangeOne (approved):

docs/infrastructure/security.md
PeterJCLaw: Hrm, I'm less sure about this one. I would expect that in many cases most people won't need server access at all and those who do will probably need `sudo` access from the get-go. There probably is a better wording here which accounts for that, however given that this section is copied from an existing one I'm more tempted to defer this and review server setups generally (see also https://github.com/srobo/ansible/issues/52 which I spotted while working on this).

Scarzy: I suspect this will depend highly on what the server is being used for. E.g. dev - not much of a concern if everyone has root, personal data handling - significant concern.
Maybe Jake's suggestion minus the word "absolutely", then it's probably easier to justify in each circumstance.

PeterJCLaw: Thanks, I'll bear these suggestions in mind and review this more generally.

PR #197, opened 2023-11-08
Declare two roles as explicit liaisons

prophile:
These aren't competition team competencies so let's make these roles explicit liaison roles rather than area owners, to Reflect Reality.


RealOrangeOne:
Mentoring also has quite a close connection to the kit team. Having it be a "liaison" is probably the wrong abstraction, but perhaps worth explicitly noting that the role requires coordination with the kit team?

WillB97 (approved):

PR #203, opened 2024-04-02
Update LED meanings for current firmware

WillB97:


docs/kit/hardware/motor-board.md
```diff
@@ -10,6 +10,8 @@
 
     The boards pictured above is an early prototype and may differ from the current boards, it is only used to display the locations of the LEDs.
 
+The motor board contains an isolation barrier to avoid ground loops. This means power must be provided to the 12V input and USB. The USB only powers up the USB-serial adapter and not the microcontroller, as such without 12V power, the USB port will appear but will respond to any communication.
```
RealOrangeOne, line 13:
```suggestion
The motor board contains an isolation barrier to avoid ground loops. This means power must be provided to the 12V input and USB. The USB only powers up the USB-serial adapter and not the microcontroller, as such without 12V power, the USB port will appear but will not respond to any communication.
```

docs/kit/hardware/servo-board.md
```diff
@@ -6,28 +6,22 @@
...
 
 - DS1 - green
     - lit when 5V is present from either the USB or UART header, this is connected to the 3.3V regulator output so shows it is functional as well.
 - DS2-9 - blue
-    - currently unused, inaccessible over USB.
+    - lit when the corresponding servo is enabled. Setting the servo to any position will enable them. These will light even when the 12V power is not present.
 - DS10 - red/blue
     - blue is lit as soon as the firmware finishes initialisation.
-    - red is lit while the I2C interrupt routine to set the servo outputs is running.
+    - red is lit while I2C communications are failing. This occurs while 12V is not present.
```
RealOrangeOne, line 21:
```suggestion
    - red is lit while I2C communications are failing. This occurs even when 12V is not present.
```

I think this is the right reading of it. Alternatively:

```suggestion
    - red is lit while I2C communications are failing. This also occurs when 12V is not present.
```

RealOrangeOne (approved):

docs/kit/hardware/servo-board.md
WillB97: No, not having 12V is the only way I have managed to get I2C to fail. It should be seen as a symptom that there is a 12V problem.

RealOrangeOne: I'm very surprised that's the only way to get it to fail, but given you've clearly tried to break it, the wording as-is sounds fine to me.

RealOrangeOne (approved):

PR #211, opened 2024-11-30
Commentator docs

RealOrangeOne:
Commentating has generally been done by only a few people, who generally _wing it_ on the day. I've decided to write down what we've learned, so we can both remember it and improve on it over time.

Unlike previous reviews, I'm happy to use this as an opportunity to review how we do things, rather than just documenting the current state.


antonnikitin97:
👀 - just noticed I had this in my todo list... I'll try and look at this this weekend 😆 .

PR #212, opened 2024-12-14
Document the process for teams which don't arrive at the Competition

PeterJCLaw:
In theory this could apply to other events as well, however it is most useful & impactful at the Competition.

Fixes https://github.com/srobo/tasks/issues/1213

### DIFF
--- a/docs/competition/teams/process-docs/team-no-show.md
+++ b/docs/competition/teams/process-docs/team-no-show.md
@@ -1,0 +1,67 @@
+# No-show Teams Process
+
+This process outlines how we detect and respond to teams who do not arrive at the competition event.
+
+This process is ultimately part of Team Support at the venue, though is likely to be delegated to another volunteer rather than run by the Team Support Area Owner.
+The person running this process is referred to as the "No-show-teams volunteer".
+
+We expect teams to have all arrived in time for the opening ceremony, so if any teams aren't present when that starts Saturday we consider them no-show and reach out to them.
+The overview of this process is:
+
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
Hopefully this captures what we discussed/agreed around https://github.com/srobo/tasks/issues/1213, though it's possible I've missed/misinterpreted/misremember things.

samjmartin04 (approved):

docs/competition/teams/process-docs/team-no-show.md
```diff
@@ -0,0 +1,67 @@
...
+!!! warning
+    This is very hard to undo once matches start, so only mark them as dropped out if the team explicitly confirm they're not coming.
+
+## Decide if and how this affects the match schedules
+
+If there are lots of dropouts and matches haven't started and we have a ready-to-go schedule for the reduced number of teams we should swap to that, to avoid lots of gaps in matches. This is only worthwhile if the number of such teams is more than two, but somewhat depends on how the match schedule looks.
+
+In general we would only change the schedule after the first round of calls on the first day of the competition.
+Changes to the schedule after the second round are expected to be too close to matches and would be disruptive.
+On later days, the schedule cannot easily be changed as doing while preserving the fairness of the matches in aggregate is very difficult.
```
samjmartin04, line 57: As an aside, this would actually be possible with my new match scheduler, not that it's usually that worth doing if it's the second day. If we found out a team has dropped out Saturday lunchtime for example, we could change it then if we really wanted to

PR #216, opened 2025-01-23
Bump to Python 3.11

PeterJCLaw:
I've been using this locally for a while, it works fine

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
Question: Why not 3.13?

PeterJCLaw:
> Question: Why not 3.13?

The README indicates `< 3.12` which suggests there's some known issues above 3.11. I didn't have time to investigate and this feels easy and worthwhile even if imperfect.

RealOrangeOne:
Following the tooling update, I think it's worth investigating bumping to the latest Python.

PR #220, opened 2025-02-19
Upgrade mkdocs

RealOrangeOne:
Closes #122 
Closes #145 

This is a big bump, but everything appears to work fine. The solution to not being able to have top-level meta pages anymore is to simply remove most of them, and move the contributing under "volunteering". We really don't need the license and project README in the final runbook.

The build also does more thorough checking of related links, so I've fixed some of those at the same time

This also fixes some rendering issues with mermaid templates, since that package can be updated.

I think the best way to test this is either blind faith or to check the repository out locally and have a click about.


RealOrangeOne:
The diff is going to be huge. The theme is upgraded, so every single file is technically being updated.

WillB97 (approved):
The changes look reasonable.
Does the large number of links that needed to be updated indicate that these links are broken in the current version or just that less extrapolation is now done by mkdocs?

docs/volunteering/glossary.md
```diff
@@ -57,11 +57,6 @@ _Mentor_:
 
 : a volunteer who helps teams build their robots
 
-_PLOD_:
-
-: an initialism of [_Pink LED of Death_](/kit/hardware/servo-board.md#pink-led-of-death-plod-bug), a
-bug that is present in the Servo Board that can be identified by a pink / purple light on the board.
-
```
PeterJCLaw, line 64: Why is this being removed here? This content change doesn't seem related to the upgrade.

docs/competition/teams/README.md
```diff
@@ -7,15 +7,13 @@ original:
 
 This is the documentation for the competition team coordinator role.
 
-There are 7 subordinate roles at the event, each of which is described
+There are 5 subordinate roles at the event, each of which is described
```
PeterJCLaw, line 10:
```suggestion
There are various subordinate roles at the event, each of which is described
```
(avoids needing to keep this number in sync)

PeterJCLaw (approved):
This seems fine. IMO it's rather unfortunate to lose the visibility on the contribution guide, but probably an ok trade-off to get back onto a more recent build tool version.

I've tested this as best I can (see prior comment). It would be good to know whether the differences I've found are those which are expected, including if there are any other key changes that I missed.

docs/volunteering/glossary.md
RealOrangeOne: The "Pink LED of Death" link doesn't exist, so it needs to go. Rather than removing the link, might as well remove the entire item, since the bug doesn't exist either.

RealOrangeOne:
> pages which don't have an h1 are no longer easily searchable by their first heading

I'm still able to find that page under both "Tickets" and "Competition preparation".

> per-page edit links are gone from pages

Restored.

> next/previous page links in the footer are gone

I can't immediately see a way of restoring these, but given our content doesn't generally follow a linear order, I don't think it's a big loss.

RealOrangeOne:
The diff of this PR is too large for the diff bot. Because the rest of CI has passed, I'm going to override and merge it.

PR #231, opened 2025-06-07
Mkdocs tweaks

RealOrangeOne:
`mkdocs-material` has a number of nice features we weren't using. Given they're there, we might as well be taking advantage of them. Specifically:

- Showing created and modified dates at the bottom of content
- Dark mode :partying_face: (uses user preferences, plus with explicit toggle)
- Search suggestions
- Restore `superfences` for syntax highlighting, as the mermaid plugin has support for this now
- Set the font to Open Sans to match our brand
- Tweak the navbar icons, particularly repository link

Also fixes showing section indexes when you click the dropdown - this was broken since a previous update. Note that the pages were still accessible, they just required an extra click to access.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -65,13 +65,11 @@ extra:
 
 # Extensions
 markdown_extensions:
-  # Note: the mermaid2 plugin also relies on code blocks for graphs, which isn't
-  # completely compatible with the following extensions, so we keep them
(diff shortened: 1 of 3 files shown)


RealOrangeOne:
(Ignore the CI failure - the diff can't be posted because it's too large).

